PR #5, opened 2020-09-24
TECH-4779: move hobo_fields to declare_schema

ColinDKelley:
The early commits here were mechanical, including `rubocop -a` and `-A`. You might want to view commit-by-commit for the first couple, so you can just skim them. Or I can break those into separate PR(s) if you want.


jebentier (commented):
Code changes look good to me, just one question. Was part of the work here to also get TravisCI hooked up for testing rather than continuing to use Jenkins?

ColinDKelley:
Yes, the ticket includes Travis:

> Set it up to run in TravisCI with Coveralls support.

I was assuming that was a separate step, but are there changes you expected to see in this PR for it?

ColinDKelley:
@jebentier Can you review what I've added here to run in Travis? That matrix isn't all green, but I created a separate ticket to green them up by moving them `rspec`: TECH-4872.

.travis.yml
```diff
@@ -0,0 +1,37 @@
...
+    - gemfile: gemfiles/rails_5.gemfile
+      rvm: 2.4.5
+    - gemfile: gemfiles/rails_6.gemfile
+      rvm: 2.4.5
+  allow_failures:
+    - rvm: ruby-head
+install:
+  - gem env
+  - gem list bundler
+  - rm -f /home/travis/.rvm/rubies/ruby-2.*/lib/ruby/gems/2.*/specifications/default/bundler-2.*.gemspec
```
ColinDKelley, line 28: Uggh! Ruby 2.7 now ships with bundler 2.0, set as a "default". We aren't ready to move to 2.0 just yet. So this hack uninstalls it so we can install 1.17.3 below and have it take precedence.

.travis.yml
```diff
@@ -0,0 +1,37 @@
...
+  - echo y | rvm @global do gem install bundler -v 1.17.3 --force --default
+  - gem install bundler -v 1.17.3 --force --default
+  - gem install bundler -v 1.17.3
+  - gem list bundler
+  - bundle --version
+  - bundle install --jobs=3 --retry=3 --path=${BUNDLE_PATH:-vendor/bundle}
```
jebentier, line 34: By moving the installation of the proper bundle version into the `before_install` section the `bundle install` step is unnecessary. It should simplify a lot of this to
```suggestion
before_install:
  - gem install bundler -v 1.17.3 --force --default
```

ColinDKelley: The `rm -f` part is necessary. But I am moving up to `before_install` and removing some of the debugging output.

PR #7, opened 2020-10-07
TECH-4876: use rails apis for getting internal metadata table names

jebentier:
This is converting from a statically defined list for ignorable tables to use the Rails APIs in order to get the configured metadata tables.

Note: `schema_info` was removed because it is the table that Rails used instead of `schema_migrations` as the default `schema_migrations_table_name` pre Rails 3.  Because (1) this gem does not support anything pre Rails 4, and (2) any projects that are still using `schema_info` instead of `schema_migrations` for the schema migration table should be setting the `schema_migrations_table_name` attribute in their Rails configuration, this static declaration has been removed.

### DIFF
--- a/lib/generators/declare_schema/migration/migrator.rb
+++ b/lib/generators/declare_schema/migration/migrator.rb
@@ -230,10 +230,10 @@ module Generators
             end
 
           [
+            'schema_info',
             ActiveRecord::Base.schema_migrations_table_name,
             ActiveRecord::Base.internal_metadata_table_name,
-            sessions_table,
-            'schema_info'
+            sessions_table
           ].compact
         end
 
--- a/lib/generators/declare_schema/migration/migrator.rb
+++ b/lib/generators/declare_schema/migration/migrator.rb
@@ -232,7 +232,8 @@ module Generators
           [
             ActiveRecord::Base.schema_migrations_table_name,
             ActiveRecord::Base.internal_metadata_table_name,
-            sessions_table
+            sessions_table,
+            'schema_info'
           ].compact
         end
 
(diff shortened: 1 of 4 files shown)


lib/generators/declare_schema/migration/migrator.rb
```diff
@@ -229,7 +229,11 @@ def always_ignore_tables
               nil
             end
 
-          ['schema_info', 'schema_migrations', 'ar_internal_metadata', sessions_table].compact
+          [
```
ColinDKelley, line 232: Didn't we say at standup that we'd keep `schema_info` in the list too, just in case? I don't think the odds are high that anyone would have a table of that name that they wanted `declare_schema` to manage, so I don't think it's very risky to leave it in the list for now, until we make all of these configurable.

jebentier: My bad, I thought we decided we could remove it. I'll add it back.

lib/generators/declare_schema/migration/migrator.rb
```diff
@@ -229,7 +229,12 @@ def always_ignore_tables
...
+            ActiveRecord::Base.schema_migrations_table_name,
+            ActiveRecord::Base.internal_metadata_table_name,
+            sessions_table,
+            'schema_info'
```
ColinDKelley, line 236: I like 'schema_info` at the top next to the new `schema_migrations_table_name`.
```suggestion
            'schema_info',
            ActiveRecord::Base.schema_migrations_table_name,
            ActiveRecord::Base.internal_metadata_table_name,
            sessions_table
```

ColinDKelley (approved):
Just one suggestion on ordering.

PR #8, opened 2020-10-09
TECH-4872/move rdoctests to rspec

ColinDKelley:


lib/generators/declare_schema/support/model.rb
```diff
@@ -4,6 +4,39 @@
 
 module DeclareSchema
   module Support
+    class IndentedBuffer
```
ColinDKelley, line 7: I got tired of fighting ERB failures on travis (mystery errors inside the ruby runtime) that didn't reproduce locally. So I tossed that ERB and wrote this tiny class to handle the generated formatting. The big plus here is that it automatically produces matching indentation.

jebentier (approved):
This looks good! I think something that we can do in the future (definitely not needed for this PR seeing how this is already taking care of so much) is that I really like the way [thoughtbot/appraisal](https://github.com/thoughtbot/appraisal/blob/master/spec/acceptance/appraisals_file_bundler_dsl_compatibility_spec.rb) does their acceptance tests. They created a very simple and easy to use DSL that implement the common patterns scattered across these files to create a more easily readable test case. I think this would be a good pattern for us to follow for these tests in the future that are testing the outward affects of the gem being used, and not so much the internal workings of the gem classes.

ColinDKelley:
merged

ColinDKelley:
@jebentier 

> I think something that we can do in the future (definitely not needed for this PR seeing how this is already taking care of so much) is that I really like the way thoughtbot/appraisal does their acceptance tests.

Sounds good! Can you make a ticket under the TECH-4525 epic?

jebentier:
Created a ticket to do the test framework enhancements

PR #13, opened 2020-10-30
Tech 4972 create callback for loading other models

jebentier:
### Added
- Added a new callback `before_generating_migrations` to the `Migrator` that can be
defined in order to custom load more models that might be missed by `eager_load!`

## Migrator Configuration

The following configuration options are available for the gem and can be used
during the initialization of your Rails application.

### before_generating_migrations callback

During the initializtion process for generating migrations, `DeclareSchema` will
trigger the `eager_load!` on the `Rails` application and all `Rails::Engine`s loaded
into scope.  This sometimes won't load all the necessary models for generating
migrations against, so we've introduced a callback that when defined is executed just
after this eager loading process.

**Example Configuration**

```ruby
DeclareSchema::Migration::Migrator.before_generating_migrations do
  require 'lib/some/hidden/models.rb'
end
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,9 +59,8 @@ during the initialization of your Rails application.
 
 During the initializtion process for generating migrations, `DeclareSchema` will
 trigger the `eager_load!` on the `Rails` application and all `Rails::Engine`s loaded
-into scope.  This sometimes won't load all the necessary models for generating
-migrations against, so we've introduced a callback that when defined is executed just
-after this eager loading process.
+into scope.  If you need to generate migrations for models that aren't automatically loaded by `eager_load!`,
+load them in the `after_load_rails_models` block.
 
 **Example Configuration**
 
--- a/README.md
+++ b/README.md
@@ -55,17 +55,17 @@ Note that the migration generator is interactive -- it can't tell the difference
 The following configuration options are available for the gem and can be used
 during the initialization of your Rails application.
 
-### after_load_rails_models callback
+### before_generating_migration callback
 
 During the initializtion process for generating migrations, `DeclareSchema` will
 trigger the `eager_load!` on the `Rails` application and all `Rails::Engine`s loaded
 into scope.  If you need to generate migrations for models that aren't automatically loaded by `eager_load!`,
-load them in the `after_load_rails_models` block.
+load them in the `before_generating_migration` block.
 
 **Example Configuration**
 
 ```ruby
-DeclareSchema::Migration::Migrator.after_load_rails_models do
+DeclareSchema::Migration::Migrator.before_generating_migration do
   require 'lib/some/hidden/models.rb'
 end
 ```
--- a/README.md
+++ b/README.md
@@ -50,6 +50,27 @@ Migration filename: [<enter>=declare_schema_migration_1|<custom_name>]: add_comp
 ```
 Note that the migration generator is interactive -- it can't tell the difference between renaming something vs. adding one thing and removing another, so sometimes it will ask you to clarify.
 
+## Migrator Configuration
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -50,6 +50,27 @@ Migration filename: [<enter>=declare_schema_migration_1|<custom_name>]: add_comp
...
+The following configuration options are available for the gem and can be used
+during the initialization of your Rails application.
+
+### after_load_rails_models callback
+
+During the initializtion process for generating migrations, `DeclareSchema` will
+trigger the `eager_load!` on the `Rails` application and all `Rails::Engine`s loaded
+into scope.  This sometimes won't load all the necessary models for generating
+migrations against, so we've introduced a callback that when defined is executed just
+after this eager loading process.
```
ColinDKelley, line 64:
```suggestion
into scope.  If you need to generate migrations for models that aren't automatically loaded by `eager_load!`,
load them in the `after_load_rails_models` block.
```

README.md
```diff
@@ -50,6 +50,27 @@ Migration filename: [<enter>=declare_schema_migration_1|<custom_name>]: add_comp
...
+
+**Example Configuration**
+
+```ruby
+DeclareSchema::Migration::Migrator.after_load_rails_models do
```
ColinDKelley, line 69: I wonder if it will be clear that this is just used for migrations? Let's try to come up with a name that captures that. And I'm not sure the order matters, so "after" might not be important. How about
```
DeclareSchema::Migration::Migrator.load_models_for_migration do
```
or maybe
```
DeclareSchema::Migration::Migrator.load_other_models_for_migration do
```
or maybe just
```
DeclareSchema::Migration::Migrator.models_for_migration do
```
?

lib/generators/declare_schema/migration/migrator.rb
```diff
@@ -110,6 +112,11 @@ def fix_native_types(types)
           def native_types
             @native_types ||= fix_native_types(connection.native_database_types)
           end
+
+          def after_load_rails_models(&blk)
+            block_given? or raise ArgumentError, 'A block is required when setting the after_load_rails_models callback'
```
ColinDKelley, line 117: Nit: Given that you have `&blk` argument, I think it's slightly more consistent to use
```
blk or raise ...
```
And why not call it `block`?

ColinDKelley (commented):
Looks good!
Just some wording/naming suggestions.

README.md
jebentier: I was thinking that this hook only being for during migrations was evident by the context of the class it's attached to as well as the documentation. I'm good with renaming it, but I don't think the `for_migration` is necessary since it's attached to the migration generator.

lib/generators/declare_schema/migration/migrator.rb
jebentier: Good point. Thank you. I slipped into old ruby habits with the naming of `blk` instead of `block`

README.md
ColinDKelley: Just because it's attached to the Migrator doesn't make it clear that it's _only_ used when generating migrations.

Here's the comment in the code today:

>   Call Fact Shards are optimized to be loaded only when needed, load these when the declare_schema migrator is running.

So how about:
```
DeclareSchema::Migration::Migrator. before_generating_migration do
```

jebentier: That naming sounds good to me, I'll push that up now

ColinDKelley (approved):
One last thought. I'm good either way. I wonder if singular might be better, since we only generate one migration at any time?
```
before_generating_migration
```

PR #14, opened 2020-10-30
STORY-8887: add support for belongs_to optional:

ColinDKelley:
### Added
- Added support for `belongs_to optional:`.
If given, it is passed through to `ActiveRecord`'s `belong_to`.
If not given in Rails 5+, the `optional:` value is set equal to the `null:` value (default: `false`) and that is passed to `ActiveRecord`'s `belong_to`.
Similarly, if `null:` is not given, it is inferred from `optional:`.
If both are given, their values are respected, even if contradictory; this is a legitimate case when migrating to/from an optional association.

### DIFF
--- a/lib/declare_schema/model.rb
+++ b/lib/declare_schema/model.rb
@@ -108,7 +108,12 @@ module DeclareSchema
       # Extend belongs_to so that it creates a FieldSpec for the foreign key
       def belongs_to(name, scope = nil, **options, &block)
         column_options = {}
-        column_options[:null] = options.delete(:null) || false
+
+        column_options[:null] = if options.has_key?(:null)
+                                  options.delete(:null)
+                                elsif options.has_key?(:optional)
+                                  options[:optional] # infer :null from :optional
+                                end || false
         column_options[:default] = options.delete(:default) if options.has_key?(:default)
         column_options[:limit] = options.delete(:limit) if options.has_key?(:limit)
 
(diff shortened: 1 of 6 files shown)


lib/declare_schema/model.rb
```diff
@@ -106,17 +106,14 @@ def rails_default_primary_key
...
-      def belongs_to(name, *args, &block)
-        if args.size == 0 || (args.size == 1 && args[0].is_a?(Proc))
-          options = {}
-          args.push(options)
-        elsif args.size == 1
-          options = args[0]
-        else
-          options = args[1]
-        end
+      def belongs_to(name, scope = nil, **options, &block)
```
ColinDKelley, line 109: Hurray for `**`!

spec/lib/declare_schema/migration_generator_spec.rb
```diff
@@ -682,5 +688,84 @@ class Ad < ActiveRecord::Base
     ActiveRecord::Migration.class_eval(up)
     expect(Ad.field_specs['company'].options[:validates].inspect).to eq("{:presence=>true, :uniqueness=>{:case_sensitive=>false}}")
   end
-end
 
+  describe 'belongs_to' do
```
ColinDKelley, line 692: This shows how broken `rdoctests` were! Since I TDD'd this, I wound up with 8-10 separate tests.

spec/lib/declare_schema/prepare_testapp.rb
```diff
@@ -18,6 +18,8 @@
 require 'rails/generators'
 Rails::Generators.configure!(Rails.application.config.generators)
 
+ActiveRecord::Base.connection.schema_cache.clear!
```
ColinDKelley, line 21: It took me more than an hour to figure out that this `schema_cache` was messing up tests by carrying state across from one test to the next.

jebentier (approved):
LGTM, nice and clean

retroGiant89 (approved):
Thanks for the comments explaining some of this :)

PR #18, opened 2020-11-02
TECH-4985: generate migrations that can run in Rails 4

ColinDKelley:
### Fixed
- Migrations are now generated where the `[4.2]` is only applied after `ActiveRecord::Migration` in Rails 5+ (since Rails 4 didn't know about that notation).


jebentier (approved):

ColinDKelley:
Merging.

PR #19, opened 2020-11-06
Tech 4937 create test dsl for acceptance tests

jebentier:
This is the pre-review PR for the new helpers that are being added for running acceptance tests in the gem. These should help with readability and understandability of the code that is doing the tests.

The following helpers have been added so far with:
1. `generate_model` acts as a shorthand for running the command line generator method for `declare_schema:model`
2. `generate_migrations` acts as a shorthand for executing the command line generator `declare_schema:migration`
3. `clean_up_model` cleans up the necessary generated files for a given model
4. `expect_model_definition_to_eq` is a slim helper to help with comparing generated model files with the expected generated  source
5. `expect_test_definition_to_eq` is a slim helper to help with comparing generated model test files with the expected generated source
6. `expect_test_fixture_to_eq` is a slim helper to help with comparing generated fixture files with the expected generated source
7. `load_models` loads the test app's models into the ruby path
8. `migrate_up` is an RSpec matcher that takes in the result of `Migrator.run` and compares its expectation against the up portion of the migrations
9. `migrate_down` is an RSpec matcher that takes in the result of `Migrator.run` and compares its expectation against the down portion of the migrations

I've gone ahead and implemented the majority of usages, but not all because I would like some eyes on the new contract to make sure it's achieving it's original purpose before finalizing anything.

### DIFF
--- a/spec/lib/declare_schema/generator_spec.rb
+++ b/spec/lib/declare_schema/generator_spec.rb
@@ -40,7 +40,7 @@ RSpec.describe 'DeclareSchema Migration Generator' do
     expect_test_fixture_to_eq('alpha/beta', <<~EOS)
       # Read about fixtures at http://api.rubyonrails.org/classes/ActiveRecord/FixtureSet.html
 
-      # This model initially had no columns defined. If you add columns to the
+      # This model initially had no columns defined.  If you add columns to the
       # model remove the '{}' from the fixture names and add the columns immediately
       # below each fixture, per the syntax in the comments below
       #
@@ -48,7 +48,7 @@ RSpec.describe 'DeclareSchema Migration Generator' do
       # column: value
       #
       two: {}
-      # column: value
+      #  column: value
     EOS
 
     $LOAD_PATH << "#{TESTAPP_PATH}/app/models"
--- a/spec/lib/declare_schema/generator_spec.rb
+++ b/spec/lib/declare_schema/generator_spec.rb
@@ -8,9 +8,7 @@ RSpec.describe 'DeclareSchema Migration Generator' do
   it "generates nested models" do
     generate_model 'alpha/beta', 'one:string', 'two:integer'
 
-    expect(File.exist?('app/models/alpha/beta.rb')).to be_truthy
-
-    expect(File.read('app/models/alpha/beta.rb')).to eq(<<~EOS)
+    expect_model_definition_to_eq('alpha/beta', <<~EOS)
       class Alpha::Beta < #{active_record_base_class}
 
         fields do
@@ -21,7 +19,7 @@ RSpec.describe 'DeclareSchema Migration Generator' do
       end
     EOS
 
-    expect(File.read('app/models/alpha.rb')).to eq(<<~EOS)
+    expect_model_definition_to_eq('alpha', <<~EOS)
       module Alpha
         def self.table_name_prefix
           'alpha_'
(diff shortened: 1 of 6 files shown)


ColinDKelley:
I like how this is looking!

spec/support/acceptance_spec_helpers.rb
```diff
@@ -0,0 +1,61 @@
...
+    expect(File.exist?(test_file_path)).to be_truthy
+    expect(File.read(test_file_path)).to eq(expectation)
+  end
+
+  def expect_test_fixture_to_eq(model, expectation)
+    fixture_file_path = "#{TESTAPP_PATH}/test/fixtures/#{model}.yml"
+
+    expect(File.exist?(fixture_file_path)).to be_truthy
+    expect(File.read(fixture_file_path)).to eq(expectation)
+  end
```
dcaddell, line 31: These expectations look the same
```
    expect(File.exist?(model_file_path)).to be_truthy
    expect(File.read(model_file_path)).to eq(expectation)
```
Any argument against factoring that part out into a different method?

yinonrousso:
Looks great!

spec/support/acceptance_spec_helpers.rb
jebentier: Absolutely. Went ahead and dried this up.

spec/lib/declare_schema/generator_spec.rb
```diff
@@ -6,44 +6,85 @@
...
+
       class Alpha::BetaTest < ActiveSupport::TestCase
         # test "the truth" do
         #   assert true
         # end
       end
     EOS
 
-    expect(File.exist?('test/fixtures/alpha/beta.yml')).to be_truthy
+    case Rails::VERSION::MAJOR
```
dcaddell, line 40: These cases all look identical. Copy/paste bug?

spec/lib/declare_schema/migration_generator_spec.rb
```diff
@@ -70,9 +73,11 @@ class Advert < ActiveRecord::Base
...
     end
 
-    up, down = migrate
-    expect(up).to eq("remove_column :adverts, :published_at")
-    expect(down).to eq("add_column :adverts, :published_at, :datetime")
+    expect(migrate).to(
+      migrate_up("remove_column :adverts, :published_at").and(
+        migrate_down("add_column :adverts, :published_at, :datetime")
+      )
+    )
```
dcaddell, line 80: It looks like we follow this `expect(x).to(migrate_up(y)).and(migrate_down(z))` pattern pretty often, and the syntax for it looks pretty easy to mess up imo. What do you think about adding a helper to take care of the syntax for you? 

```
def expect_migration_to_eq(migration, up:, down:)
   expect(migration).to(migrate_up(up)).and(migrate_down(down))
end
```

I'm torn because I like the way the current setup reads, so just wondering what your thoughts are on making it easier to write.

spec/lib/declare_schema/generator_spec.rb
jebentier: There are slight differences. Rails 6 introduces `https` to the link instead of `http`, and Rails 5 corrected some spacing in the formatting of the file.

spec/lib/declare_schema/migration_generator_spec.rb
jebentier: I played around with that, but it ends up making the contract that much messier because you cant use inline `HEREDOCS` as nicely anymore. I think this pass is good enough and makes progress and if new patterns arise, we can continue to iterate.

dcaddell (approved):
Looks great!

PR #23, opened 2020-11-13
TECH-5016: suppress belongs_to optional: in rails4

ColinDKelley:
Fixes TECH-5016

### Fixed
- When passing `belongs_to` to Rails, suppress `optional:` option in Rails 4, since it predates the introduction of that option.

### DIFF
--- a/lib/declare_schema/model.rb
+++ b/lib/declare_schema/model.rb
@@ -128,13 +128,17 @@ module DeclareSchema
 
         fk = options[:foreign_key]&.to_s || "#{name}_id"
 
-        if !options.has_key?(:optional) && Rails::VERSION::MAJOR >= 5
+        if !options.has_key?(:optional)
           options[:optional] = column_options[:null] # infer :optional from :null
         end
 
         fk_options[:dependent] = options.delete(:far_end_dependent) if options.has_key?(:far_end_dependent)
 
-        super(name, scope, options)
+        if Rails::VERSION::MAJOR >= 5
+          super(name, scope, options)
+        else
+          super(name, scope, options.except(:optional))
+        end
 
         refl = reflections[name.to_s] or raise "Couldn't find reflection #{name} in #{reflections.keys}"
         fkey = refl.foreign_key or raise "Couldn't find foreign_key for #{name} in #{refl.inspect}"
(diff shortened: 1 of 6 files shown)


lib/declare_schema/model.rb
```diff
@@ -145,7 +149,7 @@ def belongs_to(name, scope = nil, **options, &block)
           index([foreign_type, fkey], index_options) if index_options[:name] != false
         else
           index(fkey, index_options) if index_options[:name] != false
-          options[:constraint_name] = options
+          # options[:constraint_name] = options
```
ColinDKelley, line 152: This is bizarre to point `options` at itself! It was leading to infinite recursion in my test. I hope we can live without it.

yinonrousso (approved):
Looks good

lib/declare_schema/model.rb
```diff
@@ -128,13 +128,17 @@ def belongs_to(name, scope = nil, **options, &block)
...
-        if !options.has_key?(:optional) && Rails::VERSION::MAJOR >= 5
+        if !options.has_key?(:optional)
           options[:optional] = column_options[:null] # infer :optional from :null
         end
 
         fk_options[:dependent] = options.delete(:far_end_dependent) if options.has_key?(:far_end_dependent)
 
-        super(name, scope, options)
+        if Rails::VERSION::MAJOR >= 5
+          super(name, scope, options)
```
dcaddell, line 138: I don't see anything using the provided block. Can we get rid of it and just call `super` here?

dcaddell (approved):
Nice and simple change. I like the new test setup 👍 Just two questions but approved pending response.

lib/declare_schema/model.rb
dcaddell: Ha, I must've loaded the PR before you left this comment. 🔪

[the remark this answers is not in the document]
ColinDKelley: Good point. I think the odds are ~40% that some form of that line comes back, but I might as well proceed on the 60% likelihood and delete it!

lib/declare_schema/model.rb
ColinDKelley: Good point. I've removed `&block` and changed this to plain `super`.

ColinDKelley:
And BTW I just pointed `web` at this and tried generating a migration. It correctly came back with
```
Database and models match -- nothing to change
```
which is a good sign that the `options[:constraint_name] = options` line won't be missed.

PR #26, opened 2020-11-18
TECH-4975: fix primary key in sqlite

ColinDKelley:
TECH-4975

### Fixed
- Sqlite now correctly infers the PRIMARY KEY so it won't attempt to add that index again.


lib/declare_schema.rb
```diff
@@ -39,6 +39,7 @@ def to_class(type)
 require 'declare_schema/field_declaration_dsl'
 require 'declare_schema/model'
 require 'declare_schema/model/field_spec'
-require 'declare_schema/model/index_spec'
+require 'declare_schema/model/index_definition'
+require 'declare_schema/model/foreign_key_definition'
```
ColinDKelley, line 43: I renamed both of these from _spec to _definition so they wouldn't be confused with the tests that also end in _spec.
Note that field_spec should be changed like that someday. But also it should probably become column_spec or attribute_spec anyway.
Also, I split the foreign_key_definition out into its own file, as is best practice.

lib/declare_schema/model/index_settings.rb
```diff
@@ -60,9 +67,9 @@ def to_add_statement(new_table_name, existing_primary_key = nil)
           to_add_primary_key_statement(new_table_name, existing_primary_key)
         else
           r = +"add_index #{new_table_name.to_sym.inspect}, #{fields.map(&:to_sym).inspect}"
```
jebentier: Unrelated to your changes, but I just noticed that this assignment has a `+` preceding the string being assigned. Is that doing anything?

jebentier (approved):

lib/declare_schema/model/index_settings.rb
ColinDKelley: Yes, the unary plus is the best practice for making a mutable string buffer. It simply overrides the "frozen_string_literal: true" comment.
`String.new` is not good because it doesn't respect the default character set. And `"<literal>".dup` works, but it's wasteful because the double-quoted part will be a frozen string literal, which is then `dup`'d to give you a mutable copy.

ColinDKelley: Note that just below I changed from `r +=` to `r <<`. This is the one case where I think mutation is fine: accumulating into a buffer that can't be seen outside this scope until we leave and return it.

jebentier: Thank you for the explanation. That's a very interesting little bit of ruby that I didn't know about. I noticed that string mutation was happening here. I didn't know that the `+` was what was allowing that to happen.

ColinDKelley: Yeah, I learned that when I did the conversion to put the "# frozen_string_literal: true" magic comment at the top of all our files a couple years ago. (I started with `String.new`, only to step right into default character set trap.) I suppose I could send a PSA around about unary `+` if it's not widely known? I hate to encourage mutation, but if you need to accumulate a string buffer, it is the best approach.
And BTW, it's doubly helpful with string interpolation. The Ruby team made an interesting decision that "frozen_string_literal: true" should apply to interpolated strings too. They wanted consistency, but it's a bit surprising since interpolated strings can't be shared, which was the major reason for freezing. So anyway: `+` suppresses that behavior and simply gives you the unfrozen interpolated string.

jebentier: I agree on not wanting to promote mutation, I think the use cases for this are few and far enough between that a PSA isn't necessary. Maybe just an inline comment here explaining the `+` and what it's doing here would be helpful to future developers in the area?

ColinDKelley: Sounds good. Adding the comment now.

PR #27, opened 2020-11-19
TECH-4995: support serialize columns

ColinDKelley:
TECH-4995

### Added
- Fields may be declared with `serialize: true` or `serialize: <serializeable-class>`, where `<serializeable-class>` 
may be `Array` (`Array` stored as YAML) or `Hash` (`Hash` stored as YAML), (`Array` or `Hash` or any scalar value stored as JSON) or any custom serializable class. This invokes `ActiveSupport`'s `serialize` macro for that field, passing the serializable class, if given.

  Note: when `serialize:` is used, any `default:` should be given in a matching Ruby type--for example, `[]` or `{}` or `{ 'currency' => 'USD' }`--in which case the serializeable class will be used to determine the serialized default value and that will be set as the SQL default.

BTW, here's how `deploybot` changes:
```
-    required_statuses    :string, array: true, limit: 1024, default: nil, null: true # ActiveRecord stores [] as null
+    required_statuses    :string, serialize: Array, limit: 1024, default: nil, null: true # ActiveRecord stores [] as null
     # TODO: rename this column (or remove it if it's not even used) to ignore_statuses
-    blacklisted_statuses :string, array: true, limit: 1024, default: nil, null: true # github statuses to ignore during deploys
-    git_status_configs   :string, array: true, limit: 1024, default: nil, null: true # Array of GitStatusConfigs
+    blacklisted_statuses :string, serialize: Array, limit: 1024, default: nil, null: true # github statuses to ignore during deploys
+    git_status_configs   :string, serialize: Array, limit: 1024, default: nil, null: true # Array of GitStatusConfigs
     can_rollback         :boolean, default: false
 
     disabled             :boolean, default: false
@@ -27,10 +27,6 @@ class DeployTypeConfig < ActiveRecord::Base
     timestamps
   end
 
-  serialize :required_statuses, Array
-  serialize :blacklisted_statuses, Array
-  serialize :git_status_configs, Array
```

### DIFF
--- a/lib/declare_schema/model.rb
+++ b/lib/declare_schema/model.rb
@@ -194,12 +194,29 @@ module DeclareSchema
 
       def add_serialize_for_field(name, type, options)
         if (serialize_value = options.delete(:serialize))
-          type == :string || type == :text or raise ArgumentError, "serialize type must be :string or :text"
+          type == :string || type == :text or raise ArgumentError, "serialize field type must be :string or :text"
           serialize_args = Array((serialize_value unless serialize_value == true))
           serialize(name, *serialize_args)
-          if options.has_key?(:default) && !options[:default].is_a?(String) && !options[:default].nil?
-            options[:default] = options[:default].to_yaml # ActiveRecord serialization is always stored as YAML
+          if options.has_key?(:default)
+            options[:default] = serialized_default(name, serialize_value == true ? Object : serialize_value, options[:default])
           end
+        end
+      end
+
+      def serialized_default(attr_name, class_name_or_coder, default)
+        # copied from https://github.com/rails/rails/blob/7d6cb950e7c0e31c2faaed08c81743439156c9f5/activerecord/lib/active_record/attribute_methods/serialization.rb#L70-L76
+        coder = if class_name_or_coder == ::JSON
+                  ActiveRecord::Coders::JSON
+                elsif [:load, :dump].all? { |x| class_name_or_coder.respond_to?(x) }
+                  class_name_or_coder
+                else
+                  ActiveRecord::Coders::YAMLColumn.new(attr_name, class_name_or_coder)
+                end
+
+        if default == coder.load(nil)
+          nil # handle Array default: [] or Hash default: {}
+        else
+          coder.dump(default)
         end
       end
 
--- a/lib/declare_schema/model.rb
+++ b/lib/declare_schema/model.rb
@@ -193,12 +193,12 @@ module DeclareSchema
       end
 
       def add_serialize_for_field(name, type, options)
-        if (serialize_value = options.delete(:serialize))
+        if (serialize_class = options.delete(:serialize))
           type == :string || type == :text or raise ArgumentError, "serialize field type must be :string or :text"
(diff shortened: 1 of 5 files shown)


lib/declare_schema/model.rb
```diff
@@ -192,7 +192,18 @@ def add_validations_for_field(name, type, args, options)
...
       end
 
-      def add_formatting_for_field(name, type, _args)
+      def add_serialize_for_field(name, type, options)
+        if (serialize_value = options.delete(:serialize))
+          type == :string || type == :text or raise ArgumentError, "serialize type must be :string or :text"
+          serialize_args = Array((serialize_value unless serialize_value == true))
+          serialize(name, *serialize_args)
+          if options.has_key?(:default) && !options[:default].is_a?(String) && !options[:default].nil?
+            options[:default] = options[:default].to_yaml # ActiveRecord serialization is always stored as YAML
```
jebentier, line 201: I don't think this comment is fully accurate. If you pass a coder in (a class that responds to `load` and `dump`) then [Rails will use that as the serializer](https://github.com/rails/rails/blob/master/activerecord/lib/active_record/attribute_methods/serialization.rb#L41-L42) instead of their built in `YAMLColumn`. They even have a built in `JSON` coder that they use instead when you pass in the `json` gem as the coder for the column.

So for instance if I were to use the rails method like this:
```ruby
serialize :object_array, JSON
```
Rails would json serialize the column rather than yaml.  So I think we need to extract and parse the second argument to the serialize [similar to how Rails does it here](https://github.com/rails/rails/blob/master/activerecord/lib/active_record/attribute_methods/serialization.rb#L70-L76) in order to properly serialize the default.

jebentier (commented):
Code looks very clean, I just have one comment around the default serialization because I think we need to expand this a bit more.

lib/declare_schema/model.rb
ColinDKelley: Hmm, I got that from the [docs](https://api.rubyonrails.org/classes/ActiveRecord/AttributeMethods/Serialization/ClassMethods.html):

> The serialization is done through YAML. If class_name is specified, the serialized object must be of that class on assignment and retrieval.

So when it's `Array` or `Hash`, they expose that type but store in YAML.

I assumed in your example of `JSON` that they would do the same thing. But you're right, they seem to store that in JSON.

I'll update the comment.

ColinDKelley: I did a lot more than update the comment:
I added explicit test blocks for `JSON`, `Hash`, `Array`, and a custom `ValueClass` coder. And for each of those, I added tests for a variety of defaults, including the special `blank?` cases that get mapped to `NULL`. Please review the recent two commits.

jebentier (approved):
This looks great! And I like the additional work to make sure the coder is respected the same way that Rails does it.

jebentier:
Side comment on this and non-blocking, should we update the README with documentation on how to use this new setting?

ColinDKelley:
Re: documentation, I have a whole ticket in the epic to create the documentation:

TECH-5017: DeclareSchema: full documentation

We probably should do that ticket pretty soon so we have the structure in place to update the docs from then on.

PR #35, opened 2020-12-10
Tech 4971 declareschema treat integer limit 8 and bigint types identically

yinonrousso:
[JIRA](https://ringrevenue.atlassian.net/browse/TECH-4971)

## Changes
`:bigint` types should be recognized as `:integer, limit 8`

## Question
~~Should we add the same functionality for [smallint and mediumint](https://github.com/rails/rails/blob/5-2-stable/activerecord/lib/active_record/connection_adapters/abstract_mysql_adapter.rb#L828-L835)?~~ Only [bigint type](https://github.com/rails/rails/tree/master/activemodel/lib/active_model/type) was added.


lib/declare_schema/model/field_spec.rb
```diff
@@ -54,6 +53,9 @@ def initialize(model, name, type, options = {})
           end
         when :string
           @options[:limit] or raise "limit must be given for :string field #{model}##{@name}: #{@options.inspect}; do you want `limit: 255`?"
+        when :bigint
+          @type = :integer
+          @options[:limit] = 8
```
ColinDKelley, line 58: Beware of mutating the `options` passed in from the caller--they might not like that! I don't see a `dup` above, so a good way to get a copy is with `merge`:
```
@options = options.merge(limit: 8)
```

spec/lib/declare_schema/migration_generator_spec.rb
```diff
@@ -1093,4 +1093,29 @@ class Advert < active_record_base_class.constantize
       expect(base_class).to eq("(Rails::VERSION::MAJOR >= 5 ? ActiveRecord::Migration[4.2] : ActiveRecord::Migration)")
     end
   end
```
ColinDKelley, line 1095: We always want the test in the matching file. We're still adapting this hobo code to be like that. Can you create a field_spec_spec and show that `different_to?` returns false in this case?

spec/lib/declare_schema/field_spec_spec.rb
```diff
@@ -0,0 +1,67 @@
+# frozen_string_literal: true
+
+RSpec.describe 'DeclareSchema Model FieldSpec' do
+  before do
+    load File.expand_path('prepare_testapp.rb', __dir__)
+  end
```
jebentier, line 6: Nit: add a line brake between line 6 and 7

jebentier (changes requested):
The code changes look good. Could you update the Changelog?

CHANGELOG.md
```diff
@@ -73,6 +77,10 @@ using the appropriate Rails configuration attributes.
 ### Added
 - Initial version from https://github.com/Invoca/hobo_fields v4.1.0.
 
+<<<<<<< HEAD
```
jebentier, line 80: looks like some merge conflict crust was left over and needs to be removed.

jebentier, line 80: looks like some merge conflict crust was left over and needs to be removed.

jebentier (approved):
LGTM, just one comment about needing to remove some merge conflict crust

spec/lib/declare_schema/field_spec_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+
+  context 'There are no model columns to change' do
+    it '#different_to should return false for int8 == int8' do
+      class Advert < ActiveRecord::Base
+        fields do
+          price :integer, limit: 8
+        end
+      end
+
+      subject = DeclareSchema::Model::FieldSpec.new(Advert, :price, :integer, { limit: 8, null: false })
```
ColinDKelley, line 16: This is redundant with the class declaration above. At a minimum you could pull the field_spec from the class:
```ruby
subject = Advert.field_specs['price']
```
But it would be even cleaner if you could simply declare the `field_spec` here and skip the whole `ActiveRecord::Base` class (since the class is a higher-level concept; here we just want to test the `FieldSpec`).
Aha, it looks like you can, as long as you pass a `position:` option. And BTW, you don't need `{ }` around the options hash since it's the last argument.
```ruby
subject = DeclareSchema::Model::FieldSpec.new(Object, :price, :integer, limit: 8, null: false, position: 0)
```
With the above line, you can omit the `class Advert` entirely.

Same comment for the other tests below.

ColinDKelley (changes requested):
Looks good, with just one comment to simplify the field_spec_spec tests to simply operate on the `FieldSpec` classes and not pull in higher-level concepts like the `fields do` block.

ColinDKelley:
Nice, those test changes look very clean @yinonrousso.

I went ahead and made the rider change I was suggesting at standup. I didn't remove the `model` parameter since it's sometimes used by the caller. But I removed the cycle by not having FieldSpec#initialize call `model.field_specs`. That removed the need for a comment about trickery--that comment was specifically an admission of the cycle!
Also, I switched to keyword args there which are much cleaner. Notice that the `position:` wasn't really an option at all. Before that was handled with `delete` but now it's clear right in the method interface.

https://github.com/Invoca/declare_schema/pull/35/commits/4ff01fad1c83b36ad7af117b01b3e6e9631d5367

ColinDKelley:
Looks like you need to rebase this branch also @yinonrousso .

yinonrousso:
That looks a lot cleaner and makes a the method interface much easier to understand! Thanks for making that change.

yinonrousso:
@ColinDKelley I have the latest master merged in and there don't seem to be any merge conflicts. Is there still a need to rebase? Is there a preference on a rebase rather than a merge?

ColinDKelley:
No worries, a merge is fine. I had the pulldown set to `rebase` (apparently it's remembered?) and that was giving warnings.

ColinDKelley (approved):

PR #40, opened 2020-12-14
Tech 4996 support collation and character set

jebentier:
# Changelog
### Added
- Added support for configuring the character set and collation for MySQL databases
  at the global, table, and field level

# Documentation
_Note: This feature currently only works for MySQL database configurations._

With the addition of emojis, there has come a need to dynamically define the
character set and collation for individual tables and columns in the database.
With `declare_schema` this can be configured at three separate levels

### Global Configuration
The character set and collation for all tables and fields can be set at the global level
using the `Generators::DeclareSchema::Migrator.default_charset=` and
`Generators::DeclareSchema::Migrator.default_collation=` configuration methods.

For example, adding the following to your `config/initializers` directory will
turn all tables into emoji supporting tables:

**declare_schema.rb**
```ruby
# frozen_string_literal: true
Generators::DeclareSchema::Migrator.default_charset   = "utf8mb4"
Generators::DeclareSchema::Migrator.default_collation = "utf8mb4_general"
```

### Table Configuration
In order to configure a table's default character set and collation, the `charset` and
`collation` arguments can be added to the `fields` block.

For example, if you have a comments model that needs emoji support, it would look
like the following:

**app/models/comment.rb**
```ruby
# frozen_string_literal: true
class Comment < ActiveRecord::Base
  fields charset: "utf8mb4", collation: "utf8mb4_general" do
    subject :string, limit: 255
    content :text,   limit: 0xffff_ffff
  end
end
```

### Field Configuration
An if you're looking to only change the character set and collation for a single field
in the table, simply set the `charset` and `collation` configuration options on the
field definition itself.

For example, if you only want to support emojis for the content of a comment, it would
look like the following:

**app/models/comment.rb**
```ruby
# frozen_string_literal: true
class Comment < ActiveRecord::Base
  fields do
    subject :string, limit: 255
    context :text,   limit: 0xffff_ffff, charset: "utf8mb4", collation: "utf8mb4_general"
  end
end
```


README.md
```diff
@@ -70,6 +70,68 @@ DeclareSchema::Migration::Migrator.before_generating_migration do
...
+class Comment < ActiveRecord::Base
+  fields charset: "utf8mb4", collation: "utf8mb4_general" do
+    subject :string, limit: 255
+    content :text,   limit: 0xffff_ffff
+  end
+end
+```
+
+### Field Configuration
+An if you're looking to only change the character set and collation for a single field
```
ColinDKelley, line 116: If you're looking...

README.md
```diff
@@ -70,6 +70,68 @@ DeclareSchema::Migration::Migrator.before_generating_migration do
...
+character set and collation for individual tables and columns in the database.
+With `declare_schema` this can be configured at three separate levels
+
+### Global Configuration
+The character set and collation for all tables and fields can be set at the global level
+using the `Generators::DeclareSchema::Migrator.default_charset=` and
+`Generators::DeclareSchema::Migrator.default_collation=` configuration methods.
+
+For example, adding the following to your `config/initializers` directory will
+turn all tables into emoji supporting tables:
```
ColinDKelley, line 86: Let's stick with calling it "utf8mb4" rather than "emoji supporting". This applies below too.

gemfiles/rails_4.gemfile
```diff
@@ -7,6 +7,7 @@ gem "bundler", "< 2"
 gem "climate_control", "~> 0.2"
 gem "pry"
 gem "pry-byebug"
+gem "mysql2"
```
ColinDKelley, line 10: Do we need a hard dependency on `mysql2`?

lib/declare_schema/model.rb
```diff
@@ -30,6 +30,10 @@ def mix_in(base)
           inheriting_cattr_reader ignore_indexes: []
           inheriting_cattr_reader constraint_specs: []
 
+          # table_options holds optional configuration for the create_table statement
+          # supported options include :charset and :collation
+          inheriting_cattr_reader table_options: HashWithIndifferentAccess.new
```
ColinDKelley, line 35: 💯

lib/declare_schema/model/field_spec.rb
```diff
@@ -54,6 +54,9 @@ def initialize(model, name, type, options = {})
           end
         when :string
           @options[:limit] or raise "limit must be given for :string field #{model}##{@name}: #{@options.inspect}; do you want `limit: 255`?"
+        else
+          @options[:collation] and raise "collation can only given for :string and :text fields"
```
ColinDKelley, line 58: can -> may
Same below.

lib/declare_schema/model/field_spec.rb
```diff
@@ -110,35 +129,73 @@ def same_type?(col_spec)
...
       end
 
-      private
+      def same_charset_and_collation?(col_spec)
+        if type.in?([:text, :string])
+          current_settings = collation_and_charset_for_column(col_spec)
+          current_settings[:collation] == collation
+        else
+          true
+        end
```
ColinDKelley, line 175: I think it's clearer to use `||` or `&&` rather than `else true`:
```
!type.in?(text, :string]) || collation == collation_and_charset_for_column(col_spec)[:collation]
```

lib/declare_schema/model/field_spec.rb
```diff
@@ -110,35 +129,73 @@ def same_type?(col_spec)
...
+
+        if ActiveRecord::Base.connection.class.name.match?(/mysql/i)
+          database_name = ActiveRecord::Base.connection.current_database
+
+          query = <<~EOS
+            SELECT C.character_set_name, C.collation_name
+            FROM information_schema.`COLUMNS` C
+            WHERE C.table_schema = "#{database_name}" AND C.table_name = "#{table_name}" AND C.column_name = "#{column_name}";
+          EOS
+          defaults = ActiveRecord::Base.connection.select_one(query)
```
ColinDKelley, line 190: It looks like `query` is just used once. If so, why not just put it inline?
```
defaults = ActiveRecord::Base.connection.select_one(<<~EOS)
            SELECT C.character_set_name, C.collation_name
            FROM information_schema.`COLUMNS` C
            WHERE C.table_schema = "#{database_name}" AND C.table_name = "#{table_name}" AND C.column_name = "#{column_name}";
          EOS
```

ColinDKelley: Also, the connection has methods to quote the name of a table and column and I think schema. Best to use those rather than do interpolation inside quotes. In fact it's a bug if any of those names were to have a character in it that needs to be escaped.

lib/declare_schema/model/table_options_definition.rb
```diff
@@ -0,0 +1,78 @@
...
+            SELECT CCSA.character_set_name, CCSA.collation_name
+            FROM information_schema.`TABLES` T, information_schema.`COLLATION_CHARACTER_SET_APPLICABILITY` CCSA
+            WHERE CCSA.collation_name = T.table_collation AND T.table_schema = "#{connection.current_database}" AND T.table_name = "#{table_name}";
+          EOS
+          defaults = connection.select_one(query)
+
+          {
+            charset:   defaults["character_set_name"],
+            collation: defaults["collation_name"]
+          }
```
ColinDKelley, line 38: Can you document the contract for this method? Does it guarantee to have the keys `charset:` and `collation:`? Are those values guaranteed to be non-nil? If so, can you enforce that here?
```
          {
            charset:   defaults["character_set_name"] or raise "character_set_name missing from #{defaults.inspect}",
            collation: defaults["collation_name"] or raise "collation_name missing from #{defaults.inspect}",
          }
```

ColinDKelley (changes requested):

gemfiles/rails_4.gemfile
jebentier: Only for testing. This is because of the explicit checks that we're doing and testing around only supporting `charset` and `collation` initially for `MySQL` connections.

README.md
```diff
@@ -70,6 +70,68 @@ DeclareSchema::Migration::Migrator.before_generating_migration do
 end
 ```
 
+## Declaring Character Set and Collation
+_Note: This feature currently only works for MySQL database configurations._
+
+With the addition of emojis, there has come a need to dynamically define the
```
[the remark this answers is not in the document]
jebentier: Done 556ad3e16ab8af63e5f0da68975a77712a3c2f6e

README.md
jebentier: Done 556ad3e16ab8af63e5f0da68975a77712a3c2f6e

README.md
jebentier: Done 556ad3e16ab8af63e5f0da68975a77712a3c2f6e

lib/declare_schema/model/field_spec.rb
jebentier: Done 05f11dcde337c55ed4f018049fb26d837db6f78a

lib/declare_schema/model/field_spec.rb
jebentier: Done 851956a5c5a4c55ab99e69a9b461201df9c7251d

lib/declare_schema/model/field_spec.rb
jebentier: Done 851956a5c5a4c55ab99e69a9b461201df9c7251d

lib/declare_schema/model/table_options_definition.rb
jebentier: They are guaranteed to be non-nil, so I've gone ahead and updated the contract as you suggested 851956a5c5a4c55ab99e69a9b461201df9c7251d

gemfiles/rails_4.gemfile
ColinDKelley: Ah, good point--this it the GemfileI(s), not the gemspec. So no big deal.

lib/declare_schema/model/field_spec.rb
```diff
@@ -166,28 +162,33 @@ def same_attributes?(col_spec)
...
 
-      def same_charset_and_collation?(col_spec)
-        if type.in?([:text, :string])
-          current_settings = collation_and_charset_for_column(col_spec)
-          current_settings[:collation] == collation
-        else
-          true
-        end
+      def same_charset_and_collation?(table_name, col_spec)
+        current_collation_and_charset = collation_and_charset_for_column(table_name, col_spec)
```
ColinDKelley, line 166: No need for the above variable if the type isn't right, right? Assuming so, can you move this local variable and the part after the `||` below into another private method? That will keep the logic really clean.

ColinDKelley (approved):
Looks good.
Just one suggestion to create a new private method so the local variable can live there.

lib/declare_schema/model/field_spec.rb
jebentier: Good call. I went ahead and move the `type` check up and out of this method as it's out of scope for what this method's responsibilities are.

PR #48, opened 2021-01-11
Tech 4883 declareschema add and delete foreign keys using rails add foreign key and delete foreign key

yinonrousso:
[Jira Ticket](https://ringrevenue.atlassian.net/browse/TICKET-4883)

## Description
Use native rails methods to `add_foreign_key` and `remove_foreign_key`

## Summary of Changes
Added `add_foreign_key` native rails call in `ForeignKeyDefenition.to_add_statement`. Used ruby's `.inspect` method to quote strings args.


spec/lib/declare_schema/migration_generator_spec.rb
```diff
@@ -369,14 +369,14 @@ class Advert < ActiveRecord::Base
...
-        #{"execute \"ALTER TABLE adverts ADD CONSTRAINT index_adverts_on_category_id FOREIGN KEY index_adverts_on_category_id(category_id) REFERENCES categories(id) \"\n" if defined?(Mysql2)}
+        #{"add_foreign_key(\"adverts\", \"categories\", name: \"index_adverts_on_category_id\")\n" if defined?(Mysql2)}
       EOS
       .and migrate_down(<<~EOS.strip)
         remove_column :adverts, :category_id
 
         remove_index :adverts, name: :on_category_id rescue ActiveRecord::StatementInvalid
 
-        #{"remove_foreign_key('adverts', name: '')\n" if defined?(Mysql2)}
+        #{"remove_foreign_key(\"adverts\", name: nil)\n" if defined?(Mysql2)}
```
ColinDKelley, line 379: This change is surprising. Is `nil` allowed as the name? I would have expected `""` was right.

spec/lib/declare_schema/migration_generator_spec.rb
```diff
@@ -647,8 +648,8 @@ class Advert < ActiveRecord::Base
             "add_index :adverts, [:id], unique: true, name: 'PRIMARY'\n"
           elsif defined?(Mysql2)
             "execute \"ALTER TABLE adverts DROP PRIMARY KEY, ADD PRIMARY KEY (id)\"\n\n" +
-            "remove_foreign_key('adverts', name: '')\n" +
-            "remove_foreign_key('adverts', name: '')"
+            "remove_foreign_key(\"adverts\", name: nil)\n" +
+            "remove_foreign_key(\"adverts\", name: nil)"
```
ColinDKelley, line 652: Same comment.

lib/generators/declare_schema/migration/migrator.rb
```diff
@@ -564,7 +564,7 @@ def change_foreign_key_constraints(model, old_table_name)
         end
 
         def remove_foreign_key(old_table_name, fk_name)
-          "remove_foreign_key('#{old_table_name}', name: '#{fk_name}')"
+          "remove_foreign_key(#{old_table_name.inspect}, name: #{fk_name.inspect})"
```
ColinDKelley, line 567: This could be `fk_name.to_s.inspect` if we wanted to be consistent with `fk_name = nil` appearing as `name: ""`.

ColinDKelley (commented):
Just one question here, really, about treatment of `fk_name = nil` which seems to have changed.

lib/generators/declare_schema/migration/migrator.rb
yinonrousso: I was looking into this behavior as it did seem odd. Thanks, this fix looks great.

ColinDKelley (approved):

yinonrousso:
After testing the changes and only found that `remove_foreign_key` method calls use `fk.options[:constraint_name]` rather than `fk.constraint_name` which can lead the `up` and `down` migrations to not use the same foreign key name.

ColinDKelley (approved):
Nice that your testing caught that bug!

lib/declare_schema/model/foreign_key_definition.rb
```diff
@@ -51,8 +51,7 @@ def parent_table_name
       attr_writer :parent_table_name
 
       def to_add_statement(_new_table_name = nil, _existing_primary_key = nil)
```
yinonrousso, line 53: @ColinDKelley as per standup I wanted to note that these newly added args are not being used. Does it make sense to instead change this line to [fk.to_add_statement(model.table_name)](https://github.com/Invoca/declare_schema/blob/master/lib/generators/declare_schema/migration/migrator.rb#L389) to call `fk.to_add_statement` with no params since all of the data is encapsulated in the ForeignKeyDefenition object?

ColinDKelley (approved):

PR #50, opened 2021-01-14
no-jira/omit sqlite3 in mysql appraisals

ColinDKelley:
- use thoughtbot appraisal so we can use `remove_gem`
- omit `sqlite3` gem in mysql appraisals


vdinovi (approved):

ColinDKelley:
Still waiting for this appraisal gem version to be released...

PR #54, opened 2021-02-04
TECH-4814: make down migrations inverse of up

ColinDKelley:
### Changed
- Use `schema_attributes` for generating both up and down change migrations, so they are guaranteed to be symmetrical.
  Note: Rails schema dumper is still used for the down migration to replace a model that has been dropped.

### DIFF
--- a/lib/declare_schema/model/column.rb
+++ b/lib/declare_schema/model/column.rb
@@ -37,12 +37,10 @@ module DeclareSchema
         #  time: { name: "time" },
         #  date: { name: "date" },
         #  binary: { name: "blob" },
-        #  boolean: { name: "boolean" },
-        #  json: { name: "json" } }
+        #  boolean: { name: "boolean" }
         def native_types
           @native_types ||= ActiveRecord::Base.connection.native_database_types.tap do |types|
             if ActiveRecord::Base.connection.class.name.match?(/mysql/i)
-              types[:integer][:limit] ||= 11
               types[:text][:limit]    ||= 0xffff
               types[:binary][:limit]  ||= 0xffff
             end
@@ -61,6 +59,7 @@ module DeclareSchema
 
         def deserialize_default_value(column, sql_type, default_value)
           sql_type or raise ArgumentError, "must pass sql_type; got #{sql_type.inspect}"
+
           case Rails::VERSION::MAJOR
           when 4
             # TODO: Delete this code ASAP! This could be wrong, since it's using the type of the old column...which
--- a/lib/declare_schema/model/column.rb
+++ b/lib/declare_schema/model/column.rb
@@ -11,16 +11,6 @@ module DeclareSchema
           type != :primary_key && native_types.has_key?(type)
         end
 
-        def fix_native_types(types)
-          case ActiveRecord::Base.connection.class.name
-          when /mysql/i
-            types[:integer][:limit] ||= 11
-            types[:text][:limit]    ||= 0xffff
-            types[:binary][:limit]  ||= 0xffff
-          end
-          types
-        end
-
         # MySQL example:
         # { primary_key: "bigint auto_increment PRIMARY KEY",
         #   string: { name: "varchar", limit: 255 },
@@ -50,7 +40,13 @@ module DeclareSchema
         #  boolean: { name: "boolean" },
         #  json: { name: "json" } }
         def native_types
-          @native_types ||= fix_native_types(ActiveRecord::Base.connection.native_database_types)
+          @native_types ||= ActiveRecord::Base.connection.native_database_types.tap do |types|
+            if ActiveRecord::Base.connection.class.name.match?(/mysql/i)
+              types[:integer][:limit] ||= 11
+              types[:text][:limit]    ||= 0xffff
+              types[:binary][:limit]  ||= 0xffff
+            end
+          end
         end
 
         def sql_type(type)
--- a/lib/declare_schema/model/column.rb
+++ b/lib/declare_schema/model/column.rb
@@ -1,0 +1,144 @@
+# frozen_string_literal: true
+
+module DeclareSchema
+  class UnknownSqlTypeError < RuntimeError; end
+
+  module Model
+    # This class is a wrapper for the ActiveRecord::...::Column class
+    class Column
+      class << self
+        def native_type?(type)
+          type != :primary_key && native_types.has_key?(type)
+        end
+
+        def fix_native_types(types)
+          case ActiveRecord::Base.connection.class.name
+          when /mysql/i
+            types[:integer][:limit] ||= 11
+            types[:text][:limit]    ||= 0xffff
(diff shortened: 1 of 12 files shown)


lib/declare_schema/model/column.rb
```diff
@@ -0,0 +1,168 @@
+# frozen_string_literal: true
+
+module DeclareSchema
+  class UnknownSqlTypeError < RuntimeError; end
+
+  module Model
+    # This class is a wrapper for the ActiveRecord::...::Column class
```
ColinDKelley, line 7: This class is new. There was too much code in `FieldSpec` that was juggling concerns around `ActiveRecord::...Column`.

spec/lib/declare_schema/field_spec_spec.rb
```diff
@@ -1,69 +1,163 @@
...
-      when 4
-        cast_type = ActiveRecord::Type::Integer.new(limit: 8)
-        col = ActiveRecord::ConnectionAdapters::Column.new("price", nil, cast_type, "integer(8)", false)
-      else
-        sql_type_metadata = ActiveRecord::ConnectionAdapters::SqlTypeMetadata.new(sql_type: "integer(8)", type: :integer, limit: 8)
-        col = ActiveRecord::ConnectionAdapters::Column.new("price", nil, sql_type_metadata, false, "adverts")
+    describe 'integer 8' do
+      it 'returns schema attributes' do
+        subject = described_class.new(model, :price, :integer, limit: 8, null: true, position: 2)
+        expect(subject.schema_attributes(col_spec)).to eq(type: :integer, limit: 8, null: true)
```
ColinDKelley, line 31: I wound up using `compact` to throw away `nil` options, so `null: true` winds up appearing...even though it's the Rails defaults. I kind of liked being explicit, since it contradicts our convention of `null: false` in the `fields do` block.

Do you think it's helpful to have the `null: true` appear? It would only be a few lines of code if we'd rather suppress it.

jebentier: I like that `null: true` shows up in the migration. It makes it very clear to the user what the basic contract is and what will be set rather than relying on the user have in depth knowledge of mysql defaults.

ColinDKelley: Great, I'll leave it then.

lib/declare_schema/model/field_spec.rb
```diff
@@ -42,170 +56,69 @@ def initialize(model, name, type, position: 0, **options)
...
-          @options = options.merge(limit: 8)
+          @options[:limit] = 8
         end
 
-        if type.in?([:text, :string])
-          if ActiveRecord::Base.connection.class.name.match?(/mysql/i)
-            @options[:charset]   ||= model.table_options[:charset]   || Generators::DeclareSchema::Migration::Migrator.default_charset
-            @options[:collation] ||= model.table_options[:collation] || Generators::DeclareSchema::Migration::Migrator.default_collation
-          end
+        # TODO: Do we really need to support a :sql_type option? Ideally, drop it. -Colin
```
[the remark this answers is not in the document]
ColinDKelley: Might be. Or maybe it was a back door way to get finer control if you wanted a special type, like `json` before Rails 5 added support for that?

lib/generators/declare_schema/migration/migrator.rb
```diff
@@ -444,67 +431,61 @@ def change_table(model, current_table_name)
...
+              undo_changes << change_column_back(model, current_table_name, orig_col_name)
             end
-          end.compact
+          end
 
           index_changes, undo_index_changes = change_indexes(model, current_table_name, to_remove)
-          fk_changes, undo_fk_changes = if ActiveRecord::Base.connection.class.name.match?(/SQLite3Adapter/)
-                                          [[], []]
-                                        else
+          fk_changes, undo_fk_changes = if ActiveRecord::Base.connection.class.name.match?(/mysql/i)
```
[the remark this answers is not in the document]
ColinDKelley: It bugged me that we were distinguishing the pair in two different ways just a few lines apart... but I see your point that Postgres will probably wind up with foreign key support also. I'll put it back.

yinonrousso (approved):
Looks great!

jebentier (approved):

ColinDKelley:
@jebentier and @yinonrousso I had to change exceptions to warnings for invalid/missing options on fields, since `bin/regen` silently suppresses exceptions somewhere. Can you please review the changes for that?

I have a separate PR coming up for changing `web` to not cause any of those warnings.

yinonrousso (approved):
Changes look good! I'm a bit curious as to why `bin/regen` suppress exceptions. Is it just to reduce noise?

lib/declare_schema/model/field_spec.rb
```diff
@@ -83,15 +83,18 @@ def initialize(model, name, type, position: 0, **options)
         if @sql_type.in?([:string, :text, :binary, :varbinary, :integer, :enum])
           @options[:limit] ||= Column.native_types[@sql_type][:limit]
         else
-          @options.has_key?(:limit) and raise "unsupported limit: for SQL type #{@sql_type} in field #{model}##{@name}"
+          @sql_type != :decimal && @options.has_key?(:limit) and warn("unsupported limit: for SQL type #{@sql_type} in field #{model}##{@name}")
```
jebentier, line 86: One of the benefits of these being exceptions before, is that when running `rails g declare_schema:migration` it would abort the migration entirely with the exception. Is this some way that we could store the state that there are `FieldSpec` errors, and still have the migration generator abort when `FieldSpec.errors.any?`

ColinDKelley:
> I'm a bit curious as to why bin/regen suppress exceptions. Is it just to reduce noise?

Good question. I didn't have time to chase it down, but I'll see if I can find a few minutes on Friday.

Yes, it was probably to reduce noise from some exception(s) that were expected sometimes. One possible theory is that there's a `rescue nil` in there. That's a really convenient notation...that's a terrible idea. Hound knows to complain about it, but we definitely used it in our pre-hound days as a really quick way to suppress exceptions. The problem is that it suppresses all `StandardError`-derived exceptions, which is much too broad. The right way is to suppress an exact exception, or a list of exact exceptions.

lib/declare_schema/model/field_spec.rb
ColinDKelley: Certainly possible. We could even make it a configurable preference, once we have the preferences convention set up (in an on-deck ticket).

jebentier: I like that idea. If we're okay with just the warnings during migration generation for the time being and want to push the feature request of remembering the errors, and aborting the migration to a later ticket I'm okay with that. This ticket is valuable on its own and doesn't need be blocked by this.

ColinDKelley: Sounds good. I can make a ticket for that future work so we don't forget it.

PR #57, opened 2021-02-08
no-jira: support foreign key custom column:

ColinDKelley:
- Fixed a bug where the generated call to add_foreign_key() was not setting `column:`,
  so it only worked in cases where Rails could infer the foreign key by convention.
- Added missing spec for `ForeignKeyDeclaration`.


vdinovi:
Tested this generating my migration in web and it generated
```add_foreign_key("call_artifact_fragments", "call_pending_operations", column: "pending_operations_id", name: "index_call_artifact_fragments_on_pending_operations_id")```
which could be executed just fine 👍 . Thanks for the fix.

vdinovi (approved):

yinonrousso (approved):

PR #60, opened 2021-02-16
TECH-4814: fix migration generation

ColinDKelley:
### Fixed
- Exclude unknown options from FieldSpec#sql_options and #schema_attributes.
- Fixed a bug where fk_field_options were getting merged into spec_attrs after checking for equivalence,
  leading to phantom migrations with no changes, or missing migrations when just the fk_field_options changed.

### DIFF
--- a/lib/declare_schema/field_declaration_dsl.rb
+++ b/lib/declare_schema/field_declaration_dsl.rb
@@ -28,13 +28,12 @@ module DeclareSchema
       field(:lock_version, :integer, default: 1, null: false)
     end
 
-    def field(name, type, *args)
-      options = args.extract_options!
-      @model.declare_field(name, type, *(args + [@options.merge(options)]))
+    def field(name, type, *args, **options)
+      @model.declare_field(name, type, *[*args, @options.merge(options)])
     end
 
     def method_missing(name, *args)
-      field(name, args.first, *args[1..-1])
+      field(name, *args)
     end
   end
 end
--- a/lib/declare_schema/model.rb
+++ b/lib/declare_schema/model.rb
@@ -81,8 +81,7 @@ module DeclareSchema
       # arguments. The arguments are forwarded to the #field_added
       # callback, allowing custom metadata to be added to field
       # declarations.
-      def declare_field(name, type, *args)
-        options = args.extract_options!
+      def declare_field(name, type, *args, **options)
         try(:field_added, name, type, args, options)
         add_serialize_for_field(name, type, options)
         add_formatting_for_field(name, type)
--- a/lib/declare_schema/model/column.rb
+++ b/lib/declare_schema/model/column.rb
@@ -73,33 +73,33 @@ module DeclareSchema
           end
         end
 
-        # Normalizes schema attributes for the specific database adapter that is currently running
+        # Normalizes schema attributes for the given database adapter name.
         # Note that the un-normalized attributes are still useful for generating migrations because those
         # may be run with a different adapter.
-        # This method never mutates its argument. In fact it freezes it to be certain.
-        def normalize_schema_attributes(schema_attributes)
-          schema_attributes[:type] or raise ArgumentError, ":type key not found; keys: #{schema_attributes.keys.inspect}"
-          schema_attributes.freeze
-
-          case ActiveRecord::Base.connection.class.name
-          when /mysql/i
-            schema_attributes
-          when /sqlite/i
-            case schema_attributes[:type]
-            when :text
-              schema_attributes = schema_attributes.merge(limit: nil)
-            when :integer
-              schema_attributes = schema_attributes.dup
-              schema_attributes[:limit] ||= 8
-            end
-            schema_attributes
-          else
-            schema_attributes
-          end
+        # This method never mutates its argument.
+        def normalize_schema_attributes(schema_attributes, db_adapter_name)
+          case schema_attributes[:type]
+          when :boolean
+            schema_attributes.reverse_merge(limit: 1)
+          when :integer
+            schema_attributes.reverse_merge(limit: 8) if db_adapter_name.match?(/sqlite/i)
+          when :float
+            schema_attributes.except(:limit)
+          when :text
+            schema_attributes.except(:limit)          if db_adapter_name.match?(/sqlite/i)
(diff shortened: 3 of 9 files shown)


lib/declare_schema/field_declaration_dsl.rb
```diff
@@ -28,13 +28,12 @@ def optimistic_lock
       field(:lock_version, :integer, default: 1, null: false)
     end
 
-    def field(name, type, *args)
-      options = args.extract_options!
-      @model.declare_field(name, type, *(args + [@options.merge(options)]))
+    def field(name, type, *args, **options)
+      @model.declare_field(name, type, *[*args, @options.merge(options)])
```
ColinDKelley, line 32: Ah, `**` is so much nicer than `extract_options!`.

And I like this `[*` notation.

lib/declare_schema/field_declaration_dsl.rb
```diff
@@ -28,13 +28,12 @@ def optimistic_lock
...
     end
 
     def method_missing(name, *args)
-      field(name, args.first, *args[1..-1])
+      field(name, *args)
```
ColinDKelley, line 36: I can't explain why the code was pulling the array apart and splatting it in two pieces when `*` already does that. The behavior is a bit different when there are no args given, but I don't think that matters since it's not valid.

lib/declare_schema/model.rb
```diff
@@ -81,8 +81,7 @@ def ignore_index(index_name)
       # arguments. The arguments are forwarded to the #field_added
       # callback, allowing custom metadata to be added to field
       # declarations.
-      def declare_field(name, type, *args)
-        options = args.extract_options!
+      def declare_field(name, type, *args, **options)
```
ColinDKelley, line 84: Another opportunity to get rid of `extract_options!`.

lib/declare_schema/model/column.rb
```diff
@@ -73,44 +73,43 @@ def deserialize_default_value(column, sql_type, default_value)
...
+            schema_attributes.reverse_merge(limit: 8) if db_adapter_name.match?(/sqlite/i)
+          when :float
+            schema_attributes.except(:limit)
+          when :text
+            schema_attributes.except(:limit)          if db_adapter_name.match?(/sqlite/i)
+          when :datetime
+            schema_attributes.reverse_merge(precision: 0)
+          when NilClass
+            raise ArgumentError, ":type key not found; keys: #{schema_attributes.keys.inspect}"
+          end || schema_attributes
```
ColinDKelley, line 94: One big `case` on the type turned out to be more succinct. And passing the `db_adapter_name` in from the caller is less coupled, more reusable, more succinct and even slightly more efficient.

lib/declare_schema/model/column.rb
```diff
@@ -73,44 +73,43 @@ def deserialize_default_value(column, sql_type, default_value)
...
+
       def initialize(model, current_table_name, column)
         @model = model or raise ArgumentError, "must pass model"
         @current_table_name = current_table_name or raise ArgumentError, "must pass current_table_name"
         @column = column or raise ArgumentError, "must pass column"
-      end
-
-      def sql_type
-        @sql_type ||= self.class.sql_type(@column.type)
+        @sql_type = self.class.sql_type(@column.type)
```
ColinDKelley, line 112: No need for memoization since all callers need this `@sql_type`.

lib/declare_schema/model/field_spec.rb
```diff
@@ -112,14 +112,14 @@ def initialize(model, name, type, position: 0, **options)
 
         @options = Hash[@options.sort_by { |k, _v| OPTION_INDEXES[k] || 9999 }]
 
-        @sql_options = @options.except(*NON_SQL_OPTIONS)
+        @sql_options = @options.slice(*SQL_OPTIONS)
```
ColinDKelley, line 115: I first tried `& SQL_OPTIONS` but that comes from Invoca::Utils. TIL that the ActiveSupport extension of `slice` to `Hash` was actually ported over to Ruby.

lib/generators/declare_schema/migration/migrator.rb
```diff
@@ -444,12 +444,12 @@ def change_table(model, current_table_name)
             spec_attrs         = spec.schema_attributes(column)
             column_declaration = ::DeclareSchema::Model::Column.new(model, current_table_name, column)
             col_attrs          = column_declaration.schema_attributes
-            if !::DeclareSchema::Model::Column.equivalent_schema_attributes?(spec_attrs, col_attrs)
-              normalized_schema_attributes = spec_attrs.merge(fk_field_options(model, col_name_to_change))
+            normalized_schema_attrs = spec_attrs.merge(fk_field_options(model, col_name_to_change))
```
ColinDKelley, line 447: This was a subtle bug: merging the `fk_field_options` needed to happen before checking for equivalence.

yinonrousso: Nice catch!

yinonrousso (approved):
LGTM

lib/declare_schema/field_declaration_dsl.rb
jebentier: This is all so much cleaner and easier to reason about!

lib/declare_schema/field_declaration_dsl.rb
jebentier: I can wait for this to go away with TECH-5112. Do we need a `**options` here has well since `method_missing` will be receiving both `arguments` and `keyword arguments`?

ColinDKelley: If you just have `*`, that will absorb the options at the end as well.
```
def f(*args)
  args
end
=> :f
irb> f(1, 2, 3)
=> [1, 2, 3]
irb> f(1, 2, 3, a: :a, b: :b)
=> [1, 2, 3, {:a=>:a, :b=>:b}]
```

lib/generators/declare_schema/migration/migrator.rb
ColinDKelley: Hah, well it's a bug I created!

PR #61, opened 2021-02-18
Tech 5115 declareschema make foreign key default null text and string limit configurable

yinonrousso:
Create configurable defaults for:
- generate foreign key: true
- null: false
- text limit: 0xffff_ffff
- string limit: nil 

 

### Added
- Added configurable default settings for `default_text_limit`, `default_string_limit`, `default_null`,
`default_generate_foreign_keys` and `default_generate_indexing` to allow developers to adhere to project conventions.

### Changed
- Moved and deprecated default settings for `default_charset` and `default_collation` from `Generators::DeclareSchema::Migration::Migrator` to `::DeclareSchema`

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,7 +152,6 @@ Generators::DeclareSchema::Migration::Migrator.default_generate_foreign_keys = f
 The character set and collation for all tables and fields can be set at the global level
 using the `Generators::DeclareSchema::Migrator.default_charset=` and
 `Generators::DeclareSchema::Migrator.default_collation=` configuration methods.
-    - These values default to `"utf8mb4"` and `"utf8mb4_bin"` if not set.
 
 For example, adding the following to your `config/initializers` directory will
 turn all tables into `utf8mb4` supporting tables:
--- a/README.md
+++ b/README.md
@@ -70,21 +70,89 @@ DeclareSchema::Migration::Migrator.before_generating_migration do
 end
 ```
 
-## Declaring Character Set and Collation
-_Note: This feature currently only works for MySQL database configurations._
-
-MySQL originally supported UTF-8 in the range of 1-3 bytes (`mb3` or "multi-byte 3")
-which covered the full set of Unicode code points at the time: U+0000 - U+FFFF.
-But later, Unicode was extended beyond U+FFFF to make room for emojis, and with that
-UTF-8 require 1-4 bytes (`mb4` or "multi-byte 4"). With this addition, there has
-come a need to dynamically define the character set and collation for individual
-tables and columns in the database. With `declare_schema` this can be configured
-at three separate levels
-
 ### Global Configuration
+Configurations can be set at the global level to customize default declaration for the following values:
+
+#### Text Limit
+The default text limit can be set using the `Generators::DeclareSchema::Migration::Migrator.default_text_limit=` method.
+Note that a `nil` default means that there is no default-- so every declaration must be explicit.
+This will `raise` a `limit: must be provided for field :text...` error when the default value is `nil` and there is no explicit
+declaration.
+
+For example, adding the following to your `config/initializers` directory will
+set the default `text limit` value to `0xffff_ffff`:
+ 
+**declare_schema.rb**
+```ruby
+# frozen_string_literal: true
+
+Generators::DeclareSchema::Migration::Migrator.default_text_limit = 0xffff_ffff
+```
+
+#### String Limit
+The default string limit can be set using the `Generators::DeclareSchema::Migration::Migrator.default_string_limit=` method.
+Note that a `nil` default means that there is no default-- so every declaration must be explicit.
+This will `raise` a `limit: must be provided for field :string...` error when the default value is `nil` and there is no explicit
+declaration.
+
+For example, adding the following to your `config/initializers` directory will
+set the default `string limit` value to `nil`:
+ 
+**declare_schema.rb**
+```ruby
+# frozen_string_literal: true
+
+Generators::DeclareSchema::Migration::Migrator.default_string_limit = nil
+```
+
+#### Null
+The default null value can be set using the `Generators::DeclareSchema::Migration::Migrator.default_null=` method.
+Note that a `nil` default means that there is no default-- so every declaration must be explicit.
+This will `raise` a `null: must be provided for field...` error when the default value is `nil` and there is no explicit
+declaration.
+
+For example, adding the following to your `config/initializers` directory will
+set the default `null` value to `false`:
+ 
+**declare_schema.rb**
+```ruby
+# frozen_string_literal: true
+
+Generators::DeclareSchema::Migration::Migrator.default_null = false
+```
+
+#### Generate Foreign Keys
+The default value for generate foreign keys can be set using the `Generators::DeclareSchema::Migration::Migrator.default_generate_foreign_keys=` method.
+This value defaults to `true` and can only be set at the global level.
+
+For example, adding the following to your `config/initializers` directory will
+generate foreign keys:
+ 
+**declare_schema.rb**
+```ruby
+# frozen_string_literal: true
+
+Generators::DeclareSchema::Migration::Migrator.default_generate_foreign_keys = false
+```
+
+#### Generate Indexing
+The default value for generate indexing can be set using the `Generators::DeclareSchema::Migration::Migrator.default_generate_indexing=` method.
+This value defaults to `true` and can only be set at the global level.
+
+For example, adding the following to your `config/initializers` directory will
+generate indexing:
+ 
+**declare_schema.rb**
+```ruby
+# frozen_string_literal: true
+
+Generators::DeclareSchema::Migration::Migrator.default_generate_foreign_keys = false
+```
+#### Character Set and Collation
 The character set and collation for all tables and fields can be set at the global level
 using the `Generators::DeclareSchema::Migrator.default_charset=` and
 `Generators::DeclareSchema::Migrator.default_collation=` configuration methods.
+    - These values default to `"utf8mb4"` and `"utf8mb4_bin"` if not set.
 
 For example, adding the following to your `config/initializers` directory will
 turn all tables into `utf8mb4` supporting tables:
@@ -96,6 +164,17 @@ turn all tables into `utf8mb4` supporting tables:
 Generators::DeclareSchema::Migration::Migrator.default_charset   = "utf8mb4"
 Generators::DeclareSchema::Migration::Migrator.default_collation = "utf8mb4_bin"
 ```
+
+## Declaring Character Set and Collation
+_Note: This feature currently only works for MySQL database configurations._
+
+MySQL originally supported UTF-8 in the range of 1-3 bytes (`mb3` or "multi-byte 3")
+which covered the full set of Unicode code points at the time: U+0000 - U+FFFF.
+But later, Unicode was extended beyond U+FFFF to make room for emojis, and with that
+UTF-8 require 1-4 bytes (`mb4` or "multi-byte 4"). With this addition, there has
+come a need to dynamically define the character set and collation for individual
+tables and columns in the database. With `declare_schema` this can be configured
+at three separate levels
 
 ### Table Configuration
 In order to configure a table's default character set and collation, the `charset` and
@@ -136,25 +215,6 @@ class Comment < ActiveRecord::Base
 end
 ```
 
-## Addition Project Configurations
-Additional global configurations are available to match a developer's project conventions.
-Similar to setting the `character set` and `collation` at the global level you can also configure the
-default `text limit`, `string limit`, `null`, `generate foreign keys`, and `genereate indexing`.
-
-For example, adding the following to your `config/initializers` directory will set the default project configurations:
-
-**declare_schema.rb**
-```ruby
-# frozen_string_literal: true
-
-Generators::DeclareSchema::Migration::Migrator.default_text_limit            = 0xffff_ffff
-Generators::DeclareSchema::Migration::Migrator.default_string_limit          = nil
-Generators::DeclareSchema::Migration::Migrator.default_null                  = false
-Generators::DeclareSchema::Migration::Migrator.default_generate_foreign_keys = true
-Generators::DeclareSchema::Migration::Migrator.default_generate_indexing     = true
-```
-
-Note that a `nil` default for any field specification means that there is no default-- so every declaration must be explicit.
 ## Installing
 
 Install the `DeclareSchema` gem directly:
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ DeclareSchema.default_text_limit = 0xffff
 ```
 
 #### String Limit
-The default string limit can be set using the `Generators::DeclareSchema::Migration::Migrator.default_string_limit=` method.
+The default string limit can be set using the `DeclareSchema.default_string_limit=` method.
 Note that a `nil` default means that there is no default-- so every declaration must be explicit.
 This will `raise` a `limit: must be provided for field :string...` error when the default value is `nil` and there is no explicit
 declaration.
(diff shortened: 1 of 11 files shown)


lib/generators/declare_schema/migration/migrator.rb
```diff
@@ -96,6 +105,31 @@ def default_collation=(collation)
...
+
+          def default_null=(null)
+            [true, false].include? null or raise ArgumentError, "null must be either true, false, or nil (got #{null.inspect})"
+            @default_null = null
+          end
+
+          def default_generate_foreign_keys=(generate_foreign_keys)
+            [true, false].include? generate_foreign_keys or raise ArgumentError, "generate_foreign_keys must be either true or false (got #{generate_foreign_keys.inspect})"
+            @default_generate_foreign_keys = generate_foreign_keys
+            toggle_disable_constraints
```
yinonrousso, line 126: TODO: fuse these two instance variables as they do the same thing.

lib/generators/declare_schema/migration/migrator.rb
```diff
@@ -477,7 +511,7 @@ def change_table(model, current_table_name)
         end
 
         def change_indexes(model, old_table_name, to_remove)
-          Migrator.disable_constraints and return [[], []]
+          Migrator.disable_indexing and return [[], []]
```
yinonrousso, line 514: Should we a global config option to disable indexing? It seems to follow an almost identical pattern to disable fk constraints.

lib/declare_schema/model/field_spec.rb
```diff
@@ -60,18 +60,18 @@ def initialize(model, name, type, position: 0, **options)
...
           if self.class.mysql_text_limits?
             @options[:default].nil? or raise MysqlTextMayNotHaveDefault, "when using MySQL, non-nil default may not be given for :text field #{model}##{@name}"
-            @options[:limit] = self.class.round_up_mysql_text_limit(@options[:limit] || MYSQL_LONGTEXT_LIMIT)
+            @options[:limit] = self.class.round_up_mysql_text_limit(@options[:limit] || Generators::DeclareSchema::Migration::Migrator.default_text_limit)
           else
             @options[:limit] = nil
           end
         when :string
-          @options[:limit] or raise "limit: must be given for :string field #{model}##{@name}: #{@options.inspect}; do you want `limit: 255`?"
+          @options[:limit] ||= Generators::DeclareSchema::Migration::Migrator.default_string_limit or raise "limit: must be given for :string field #{model}##{@name}: #{@options.inspect}; do you want `limit: 255`?"
```
yinonrousso, line 74: A little confused on allowing default `nil`. Should we allow default to be `nil` for 2-4 mentioned in the [ticket](https://ringrevenue.atlassian.net/browse/TECH-5115) and raise if not explicitly set in field spec declaration.

lib/generators/declare_schema/migration/migrator.rb
ColinDKelley: Yes, good idea!

lib/declare_schema/model/field_spec.rb
ColinDKelley: Right, for each of those, if it has a `nil` default, raise an exception if it's not explicitly set in the the field spec options, saying a value needs to be provided (because the default is configured `nil`).

README.md
```diff
@@ -136,6 +136,25 @@ class Comment < ActiveRecord::Base
 end
 ```
 
+## Addition Project Configurations
+Additional global configurations are available to match a developer's project conventions.
+Similar to setting the `character set` and `collation` at the global level you can also configure the
+default `text limit`, `string limit`, `null`, `generate foreign keys`, and `genereate indexing`.
```
ColinDKelley, line 142: I think it will be simpler to understand if these options are just grouped with the character set ones in a "Configuration" section. No need for one to reference the other.

ColinDKelley: Please make a list (could be a ### section) that lists them out and has room for a sentence or two about each. I'd just repeat the rule about what `nil` means for the 3 cases where it's supported. Ideally, document the exception that will be raised if they break that rule.

lib/declare_schema/extensions/active_record/fields_declaration.rb
```diff
@@ -15,7 +15,7 @@ def fields(table_options = {}, &block)
       @table_options        = table_options
 
       if block
-        dsl = DeclareSchema::FieldDeclarationDsl.new(self, null: false)
+        dsl = DeclareSchema::FieldDeclarationDsl.new(self)
```
ColinDKelley, line 18: I'd forgotten that previously this could be overridden on a per-model basis. Oh well, we never found that helpful.

lib/declare_schema/model/field_spec.rb
```diff
@@ -60,18 +60,22 @@ def initialize(model, name, type, position: 0, **options)
...
+        @options[:null].nil? and raise "null: must be provided for field #{model}##{@name}: #{@options.inspect} since Generators::DeclareSchema::Migration::Migrator#default_null is set to 'nil'; do you want `null: false`?"
 
         case type
         when :text
           if self.class.mysql_text_limits?
             @options[:default].nil? or raise MysqlTextMayNotHaveDefault, "when using MySQL, non-nil default may not be given for :text field #{model}##{@name}"
-            @options[:limit] = self.class.round_up_mysql_text_limit(@options[:limit] || MYSQL_LONGTEXT_LIMIT)
+            @options[:limit] = self.class.round_up_mysql_text_limit(
+              @options[:limit] ||
+              Generators::DeclareSchema::Migration::Migrator.default_text_limit ||
```
ColinDKelley, line 72: That last || should be `or` since it's flow-of-controll.

lib/declare_schema/model/field_spec.rb
```diff
@@ -60,18 +60,22 @@ def initialize(model, name, type, position: 0, **options)
         @position = position
         @options = options.dup
 
-        @options.has_key?(:null) or @options[:null] = false
+        @options.has_key?(:null) or @options[:null] = Generators::DeclareSchema::Migration::Migrator.default_null
+        @options[:null].nil? and raise "null: must be provided for field #{model}##{@name}: #{@options.inspect} since Generators::DeclareSchema::Migration::Migrator#default_null is set to 'nil'; do you want `null: false`?"
```
ColinDKelley, line 64: Oh, I just noticed that the config is in the `Generators` namespace. I don't think it should be. (I hope to fix that someday and not use that namespace at all.)
Since it's in the gem ::DeclareSchema, the config should live there too:
```
DeclareSchema.default_null = false
etc
```

lib/declare_schema/model/field_spec.rb
ColinDKelley: This last exception should also mention the config variable to tune.

lib/declare_schema/model/field_spec.rb
```diff
@@ -107,7 +111,7 @@ def initialize(model, name, type, position: 0, **options)
           end
         else
           @options[:charset]   and warn("charset may only given for :string and :text fields for SQL type #{@sql_type} in field #{model}##{@name}")
-          @options[:collation] and warne("collation may only given for :string and :text fields for SQL type #{@sql_type} in field #{model}##{@name}")
```
ColinDKelley, line 110: Hah, that typo was in the mainline code!?

lib/generators/declare_schema/migration/migrator.rb
```diff
@@ -72,19 +72,30 @@ def constraint_specs
...
         @ignore_tables                        = []
         @before_generating_migration_callback = nil
         @active_record_class                  = ActiveRecord::Base
         @default_charset                      = DEFAULT_CHARSET
         @default_collation                    = DEFAULT_COLLATION
+        @default_text_limit                   = DEFAULT_TEXT_LIMIT
+        @default_string_limit                 = DEFAULT_STRING_LIMIT
+        @default_null                         = DEFAULT_NULL
+        @default_generate_foreign_keys        = DEFAULT_GENERATE_FOREIGN_KEYS
+        @default_generate_indexing            = DEFAULT_GENERATE_INDEXING
```
ColinDKelley, line 93: Does anyone use these DEFAULT_ constants outside of here? If not, I'd just get rid of them and put the defaults right there, inline. It will be very clear what they're for.

lib/generators/declare_schema/migration/migrator.rb
```diff
@@ -96,6 +107,31 @@ def default_collation=(collation)
...
+            @default_text_limit = text_limit
+          end
+
+          def default_string_limit=(string_limit)
+            string_limit.nil? or string_limit.is_a?(Integer) or raise ArgumentError, "string limit must be an integer or nil (got #{string_limit.inspect})"
+            @default_string_limit = string_limit
+          end
+
+          def default_null=(null)
+            [true, false, nil].include? null or raise ArgumentError, "null must be either true, false, or nil (got #{null.inspect})"
```
ColinDKelley, line 121: We use parens these days, so that'd be `include?(null)`.

Ditto below in a few places.

spec/lib/declare_schema/field_spec_spec.rb
```diff
@@ -55,6 +55,15 @@
           expect(subject.schema_attributes(col_spec)).to eq(type: :string, limit: 100, null: true)
         end
       end
+
+      it 'raises error when DEFAULT_STRING_LIMIT option is nil when not explicitly set in field spec' do
```
ColinDKelley, line 59: I don't think `DEFAULT_STRING_LIMIT` should be capitalized here. It makes it seem like a constant.

spec/lib/generators/declare_schema/migration/migrator_spec.rb
```diff
@@ -57,6 +57,76 @@ module Migration
...
+            it { should eq(true) }
+          end
+
+          context 'when explicitly set' do
+            before { described_class.default_generate_indexing = false }
+            after  { described_class.default_generate_indexing = described_class::DEFAULT_GENERATE_INDEXING }
+            it     { should eq(false) }
+          end
+        end
+
```
ColinDKelley, line 129: 💯 Tests look good!

ColinDKelley (commented):
Just a couple comments about the documentation and what namespace the configurations live in.

README.md
```diff
@@ -70,18 +70,85 @@ DeclareSchema::Migration::Migrator.before_generating_migration do
 end
 ```
 
-## Declaring Character Set and Collation
-_Note: This feature currently only works for MySQL database configurations._
+### Global Configuration
```
yinonrousso, line 73: I went ahead and moved this under the `##Migrator Configuration` heading

README.md
```diff
@@ -70,18 +70,85 @@ DeclareSchema::Migration::Migrator.before_generating_migration do
...
+This value defaults to `true` and can only be set at the global level.
+
+For example, adding the following to your `config/initializers` directory will
+generate indexing:
+ 
+**declare_schema.rb**
+```ruby
+# frozen_string_literal: true
+
+Generators::DeclareSchema::Migration::Migrator.default_generate_foreign_keys = false
```
ColinDKelley, line 149: I think this is a copy/paste error? But also, like above, I'd make the example show how to configure different from the default. I think the default here is `true` for generating indexes.

lib/declare_schema/model/field_spec.rb
ColinDKelley: ^ Does the above sound reasonable? I think it's surprising for the `DeclareSchema` gem to have any code or variables outside of that namespace. I hope we can remove the `Generators` reference entirely at some point I tried once but timed out after 20 minutes.

lib/declare_schema/model/field_spec.rb
```diff
@@ -104,7 +107,7 @@ def initialize(model, name, type, position: 0, **options)
           end
         else
           @options[:charset]   and warn("charset may only given for :string and :text fields for SQL type #{@type} in field #{model}##{@name}")
-          @options[:collation] and warne("collation may only given for :string and :text fields for SQL type #{@type} in field #{model}##{@name}")
+          @options[:collation] and warn("collation may only given for :string and :text fields for SQL type #{@type} in field #{model}##{@name}")
```
ColinDKelley, line 110: Doh! Don't do a `git blame` on that typo!

README.md
```diff
@@ -70,18 +70,85 @@ DeclareSchema::Migration::Migrator.before_generating_migration do
...
-at three separate levels
+#### Text Limit
+The default text limit can be set using the `Generators::DeclareSchema::Migration::Migrator.default_text_limit=` method.
+Note that a `nil` default means that there is no default-- so every declaration must be explicit.
+This will `raise` a `limit: must be provided for field :text...` error when the default value is `nil` and there is no explicit
+declaration.
 
-### Global Configuration
+For example, adding the following to your `config/initializers` directory will
+set the default `text limit` value to `0xffff_ffff`:
```
ColinDKelley, line 83: Similar to comments below, can you make this example something other than the default default? For example, `0xffff` would be good.

lib/generators/declare_schema/migration/migrator.rb
```diff
@@ -33,6 +36,31 @@ def default_collation=(collation)
...
+            @default_text_limit = text_limit
+          end
+
+          def default_string_limit=(string_limit)
+            string_limit.nil? or string_limit.is_a?(Integer) or raise ArgumentError, "string limit must be an integer or nil (got #{string_limit.inspect})"
+            @default_string_limit = string_limit
+          end
+
+          def default_null=(null)
+            [true, false, nil].include?(null) or raise ArgumentError, "null must be either true, false, or nil (got #{null.inspect})"
```
ColinDKelley, line 50: How about
```
null.in?([true, false, nil])
```
It's equivalent, but I think it's more clear to start with the variable you're checking. Same comments below with `generate_foreign_keys` and `generate_indexing`.

lib/declare_schema/model/field_spec.rb
yinonrousso: Agreed, I think everything should live within the `DeclareSchema` namespace.

yinonrousso: It looks as though the `Generators` namespace was reflecting the `ActiveRecord::Generators::Migration` and `ActiveRecord::Generators::Model` structure. But even so the pattern should follow as `DeclareSchema::Generators::Migration` not `Generators::DeclareSchema::Migration`.

ColinDKelley: When I tried moving from `Generators::DeclareSchema` to `DeclareSchema::Generators`, I got hung up because Rails complained; it was expecting the former. I didn't have time to track down if there's a way to tune Rails to not expect that. But in any case, let's go with this:
1. [this PR] Move the tunable settings (part of the external interface) into `::DeclareSchema`.
2. [later] Move everything into the `::DeclareSchema` namespace, with the minimal hook for Rails to find it. The cool thing in Ruby is that you can alias one constant to another with simple assignment, so all it will take is this:
```
module Generators
  module DeclareSchema
    Migration = :: DeclareSchema::Generators::Migration
  end
end
```

yinonrousso: I went ahead and moved the tunable setting for [default_text_limit](https://github.com/Invoca/declare_schema/pull/61/commits/78d103440c452ed9dd0976ddc531274eaa9fa563) in this commit. The same pattern should follow for the remaining settings.

yinonrousso: I'm also wondering whether or not there is a need to deprecate the `character set` and `collation` settings since they already exist in previous release version tunable in the `Generators::DeclareSchema::Migration::Migrator` namespace

ColinDKelley: Ideally, yes. Can you make those methods that delegate to the new location? I think that's just:
```
module Generators
  module DeclareSchema
    def default_character_set=(character_set)
      ::DeclareSchema.default_character_set = character_set
    end
    deprecate :default_character_set=, ...
  end
end
```

ColinDKelley (approved):
Looks good. 👍
Just one comment about a space.

PR #62, opened 2021-02-18
TECH-4814: remove sql_type

ColinDKelley:
I decided to add this as a rider when I realized we'd been wasting time with the `sql_type` that wasn't significantly different from `type`. I'm guessing it might have been left over from Hobo Fields supporting custom field types like `Email` that were really just a `:string` field with a different validator.

### Removed
- Removed `sql_type` that was confusing because it was actually the same as `type` (ex: :string) and not
  in fact the SQL type (ex: ``varchar(255)'`).

### DIFF
--- a/lib/declare_schema/model/column.rb
+++ b/lib/declare_schema/model/column.rb
@@ -44,6 +44,8 @@ module DeclareSchema
             if ActiveRecord::Base.connection.class.name.match?(/mysql/i)
               types[:text][:limit]    ||= 0xffff
               types[:binary][:limit]  ||= 0xffff
+
+              types[:varbinary] ||= { name: "varbinary" } # TODO: :varbinary is an Invoca addition to Rails; make it a configurable option
             end
           end
         end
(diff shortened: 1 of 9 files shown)


lib/declare_schema/model/column.rb
```diff
@@ -103,13 +95,14 @@ def equivalent_schema_attributes?(schema_attributes_lhs, schema_attributes_rhs)
...
-      attr_reader :sql_type
+      attr_reader :type
 
       def initialize(model, current_table_name, column)
         @model = model or raise ArgumentError, "must pass model"
         @current_table_name = current_table_name or raise ArgumentError, "must pass current_table_name"
         @column = column or raise ArgumentError, "must pass column"
-        @sql_type = self.class.sql_type(@column.type)
+        @type = @column.type
+        self.class.native_type?(@type) or raise UnknownTypeError, "#{@type.inspect}"
```
jebentier, line 105: This is a much cleaner contract! 👏

jebentier (approved):
This is great cleanup and was definitely left over from the custom fields provided by hobofields that are no longer implemented.

PR #63, opened 2021-02-18
TECH-5116: remove assumption that primary key is `id`

ColinDKelley:


lib/declare_schema/model/foreign_key_definition.rb
```diff
@@ -15,10 +17,10 @@ def initialize(model, foreign_key, options = {})
         @child_table = model.table_name # unless a table rename, which would happen when a class is renamed??
         @parent_table_name = options[:parent_table]&.to_s
         @foreign_key_name = options[:foreign_key]&.to_s || @foreign_key
-        @index_name = options[:index_name]&.to_s || model.connection.index_name(model.table_name, column: @foreign_key_name)
```
ColinDKelley, line 18: This `@index_name` was never used except in the lines below. So I just folded it in there.

lib/declare_schema/model/foreign_key_definition.rb
```diff
@@ -15,10 +17,10 @@ def initialize(model, foreign_key, options = {})
...
 
-        # Empty constraint lets mysql generate the name
-        @constraint_name = options[:constraint_name]&.to_s || @index_name&.to_s || ''
+        @constraint_name = options[:constraint_name]&.to_s ||
+                             options[:index_name]&.to_s ||
+                             IndexDefinition.index_name(@foreign_key_name)
```
ColinDKelley, line 23: The `index_name` method from Rails just generated an index name like `index_users_on_first_name_and_last_name` and we cut down on that verbosity because it was blowing out the MySQL length limit sometimes. (And there's no need to put the table name up front since indexes name scope is always within a table.) We were `gsub` ing before to simplify that, so I decided just to make a class method out of what we really want.

lib/declare_schema/model/habtm_model_shim.rb
```diff
@@ -0,0 +1,75 @@
+# frozen_string_literal: true
+
+module DeclareSchema
+  module Model
+    class HabtmModelShim
```
ColinDKelley, line 5: This class was wedged over in the migrator. It was time to move it out, and write a spec for it.

lib/declare_schema/model/index_definition.rb
```diff
@@ -60,6 +60,10 @@ def for_model(model, old_table_name = nil)
           index_definitions
         end
 
+        def index_name(columns)
+          "on_#{Array(columns).join("_and_")}"
```
ColinDKelley, line 64: No need to ask Rails to make the name and then `gsub` it.

lib/declare_schema/model.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'rails'
```
jebentier, line 3: Do we need to require all of `rails` here? Or is it only `activerecord` that we need to require?

lib/declare_schema/model.rb
```diff
@@ -99,8 +101,29 @@ def index_definitions_with_primary_key
         end
       end
 
-      def primary_key
-        super || 'id'
+      if ::Rails::VERSION::MAJOR < 5
+        def primary_key
+          super || 'id'
```
jebentier, line 106: Does this trip up if the model is using a composite primary key? To my understanding, ActiveRecord will warn about the composite key being used and return `nil`. I don't think we want to default to `id` in that case, but respect the composite primary key that was defined.

lib/declare_schema/model/field_spec.rb
```diff
@@ -47,11 +47,9 @@ def round_up_mysql_text_limit(limit)
...
 
       def initialize(model, name, type, position: 0, **options)
-        # TODO: TECH-5116
-        # Invoca change - searching for the primary key was causing an additional database read on every model load.  Assume
-        # "id" which works for invoca.
-        # raise ArgumentError, "you cannot provide a field spec for the primary key" if name == model.primary_key
-        name == "id" and raise ArgumentError, "you cannot provide a field spec for the primary key"
+        defined_primary_key = model.defined_primary_key
+
+        name.to_s == defined_primary_key and raise ArgumentError, "you may not provide a field spec for the primary key #{name.inspect}"
```
jebentier, line 52: There is a subtlety here that I think needs to be addressed with this check that has to do with how the `create_table` is generated. Primarily create table has the [following contract](https://github.com/rails/rails/blob/main/activerecord/lib/active_record/connection_adapters/abstract/schema_statements.rb#L154-L296). The key part being the slight difference in defining the primary key for the table. You can do one of the following two things:

1. Use `id` as the primary key and define the specific type to use in the `create_table` statement directly (we are doing this already to make `id` a `:bigint`
```ruby
create_table(:some_table, id: :bigint) do; end

CREATE TABLE some_table (
    id bigint DEFAULT NULL auto_increment PRIMARY KEY
)
```

2. Use a different column as the primary key and let the default rails type take over (in 4.2 this is int(11))
```ruby
create_table(:some_table, primary_key: 'guid') do; end

CREATE TABLE suppliers (
    guid int(11) DEFAULT NULL auto_increment PRIMARY KEY
)
```

3. Use a different column(s) as the primary key which are defined within the field specs for usage of custom type
```ruby
create_table(:some_table, primary_key: 'name') do |t|
  t.string :name
end

CREATE TABLE suppliers (
    name string DEFAULT NULL PRIMARY KEY
)
```

I think we need to make a choice right now to either support all, or block one/two of these until we get the support added.

lib/declare_schema/model.rb
ColinDKelley: It's there so we can check the `Rails::VERSION`. Other parts of this repo do the same thing, so this isn't add another dependency. It's just allowing this test to run in isolation.

ColinDKelley: Once we drop support for Rails 4, we can remove all these special cases for < 5.

lib/declare_schema/model.rb
ColinDKelley: The issue is that the `primary_key` method is built into Rails. And Rails checks for it being `nil`...sometimes raising an exception. I don't think it's acceptable for `declare_schema` to mess with Rails like that ... except this grandfather code that we needed to get Rails 4 to work for the next month or so.

lib/declare_schema/model/field_spec.rb
ColinDKelley: I agree with that background, but I'm not sure we have to do anything right now. This code is equivalent to what we've always had, right?

ColinDKelley: (BTW it's an eternal headache in this code and SQL itself that primary keys are columns and also indexes. 99% of the time a single column, but 1% they are compound. So sometimes the PK is thought of as an attribute of the table, sometimes it's a boolean option on a column, and sometimes it's a special index. It would be great to standardize on one approach. The middle option there doesn't work for compound PKs--because you need ordering too--so it's not a great choice. The latter is probably the most general and flexible. In `declare_schema` we do that some, with a convention of the index name being PRIMARY.)

jebentier: True the contract isn't changing. This would be an enhancement so it's not necessary right now.

lib/declare_schema/model.rb
jebentier: Gotcha! Sounds good

lib/declare_schema/model.rb
jebentier: Sounds good, thanks for the clarification.

jebentier (approved):

yinonrousso (approved):

jebentier (approved):
LGTM

PR #65, opened 2021-02-23
Tech 5112 create declare schema block dsl

dcaddell:
### Deprecated
- The `fields` dsl method is being deprecated.

### Added
- Added the `declare_schema` method to replace `fields`. We now expect a column's type to come before the name
i.e. `declare schema { string :title }`. Otherwise, there is no difference between `fields` and `declare_schema`.

### DIFF
--- a/.github/workflows/declare_schema_build.yml
+++ b/.github/workflows/declare_schema_build.yml
@@ -57,4 +57,4 @@ jobs:
           git config --global user.email "dummy@example.com"
           git config --global user.name "dummy"
           MYSQL_PORT=3306 bundle exec rake test:prepare_testapp[force]
-          bundle exec rake test:all < test_responses.txt
+          bundle exec rake test:all
--- a/lib/declare_schema/dsl.rb
+++ b/lib/declare_schema/dsl.rb
@@ -28,8 +28,7 @@ module DeclareSchema
       field(:lock_version, :integer, default: 1, null: false)
     end
 
-    def field(name, type, *args)
-      options = args.extract_options!
+    def field(name, type, *args, **options)
       @model.declare_field(name, type, *(args + [@options.merge(options)]))
     end
 
--- a/lib/declare_schema/dsl.rb
+++ b/lib/declare_schema/dsl.rb
@@ -1,0 +1,40 @@
+# frozen_string_literal: true
+
+require 'active_support/proxy_object'
+
+module DeclareSchema
+  class Dsl < BasicObject # avoid Object because that gets extended by lots of gems
+    include ::Kernel      # but we need the basic class methods
+
+    instance_methods.each do |m|
+      unless m.to_s.starts_with?('__') || m.in?([:object_id, :instance_eval])
+        undef_method(m)
+      end
+    end
+
+    def initialize(model, options = {})
+      @model = model
+      @options = options
+    end
+
+    attr_reader :model
+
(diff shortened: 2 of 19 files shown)


spec/lib/declare_schema/interactive_primary_key_spec.rb
```diff
@@ -11,49 +11,101 @@
...
+        end
+        self.primary_key = "id"
       end
-      self.primary_key = "id"
-    end
 
-    puts "\n\e[45m Please enter 'id' (no quotes) at the next prompt \e[0m"
-    generate_migrations '-n', '-m'
-    expect(Foo._defined_primary_key).to eq('id')
+      allow_any_instance_of(DeclareSchema::Support::ThorShell).to receive(:ask).with(/one of the rename choices or press enter to keep/) { 'id' }
```
yinonrousso, line 33: Nicely done!

lib/declare_schema/dsl.rb
```diff
@@ -0,0 +1,40 @@
...
+      field(:created_at, :datetime, null: true)
+      field(:updated_at, :datetime, null: true)
+    end
+
+    def optimistic_lock
+      field(:lock_version, :integer, default: 1, null: false)
+    end
+
+    def field(name, type, *args)
+      options = args.extract_options!
```
ColinDKelley, line 32: I think we can get rid of `extract_options!` here:
```
def field(name, type, *args, **options)
```

ColinDKelley (approved):
Looks good! Just one suggestion to remove a line.

lib/generators/declare_schema/support/model.rb
```diff
@@ -69,11 +69,11 @@ def inject_declare_schema_code_into_model_file
             def declare_model_fields_and_associations
               buffer = ::DeclareSchema::Support::IndentedBuffer.new(indent: 2)
               buffer.newline!
-              buffer << 'fields do'
+              buffer << 'declare_schema do'
               buffer.indent! do
                 field_attributes.each do |attribute|
-                  decl = "%-#{max_attribute_length}s" % attribute.name + ' ' +
-                    attribute.type.to_sym.inspect +
+                  decl = "%-#{max_attribute_length}s" % attribute.type + ' ' +
```
yinonrousso, line 75: Interesting to see how this column and type switch works!

yinonrousso (approved):
Changes look great!

yinonrousso:
One last minute change is to update the [git workflow](https://github.com/Invoca/declare_schema/blob/master/.github/workflows/declare_schema_build.yml#L60) to remove the `test_responses.txt`

.github/workflows/declare_schema_build.yml
```diff
@@ -57,4 +57,4 @@ jobs:
           git config --global user.email "dummy@example.com"
           git config --global user.name "dummy"
           MYSQL_PORT=3306 bundle exec rake test:prepare_testapp[force]
-          bundle exec rake test:all < test_responses.txt
+          bundle exec rake test:all
```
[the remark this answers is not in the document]
dcaddell: @yinonrousso is this the change that you were asking for?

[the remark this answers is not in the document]
yinonrousso: @dcaddell whoops looks like I missed that! Yes it was 👍

PR #68, opened 2021-03-17
TECH-5338: sort migration commands to meet dependencies

ColinDKelley:
### Fixed
- Migration steps are now generated in a defined dependency order, so that--for example--indexes that depend
 on columns are deleted first, before the columns themselves are deleted (since the latter implicitly does the former, which could break the migration when run). Similarly foreign keys depended on indexes.
- Related to the above, down migration steps are now always generated in exactly the reverse order of the up migration steps.

### DIFF
--- a/lib/declare_schema/schema_change/all.rb
+++ b/lib/declare_schema/schema_change/all.rb
@@ -2,6 +2,8 @@
 
 module DeclareSchema
   module SchemaChange
+    module All
+    end
     autoload :Base, 'declare_schema/schema_change/base'
     autoload :ColumnAdd, 'declare_schema/schema_change/column_add'
     autoload :ColumnChange, 'declare_schema/schema_change/column_change'
--- a/lib/declare_schema/schema_change/all.rb
+++ b/lib/declare_schema/schema_change/all.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+module DeclareSchema
+  module SchemaChange
+    autoload :Base, 'declare_schema/schema_change/base'
+    autoload :ColumnAdd, 'declare_schema/schema_change/column_add'
+    autoload :ColumnChange, 'declare_schema/schema_change/column_change'
+    autoload :ColumnRename, 'declare_schema/schema_change/column_rename'
+    autoload :ForeignKeyAdd, 'declare_schema/schema_change/foreign_key_add'
+    autoload :ForeignKeyRemove, 'declare_schema/schema_change/foreign_key_remove'
+    autoload :IndexAdd, 'declare_schema/schema_change/index_add'
+    autoload :IndexRemove, 'declare_schema/schema_change/index_remove'
+    autoload :PrimaryKeyChange, 'declare_schema/schema_change/primary_key_change'
+    autoload :TableAdd, 'declare_schema/schema_change/table_add'
+    autoload :TableChange, 'declare_schema/schema_change/table_change'
+    autoload :TableRemove, 'declare_schema/schema_change/table_remove'
+    autoload :TableRename, 'declare_schema/schema_change/table_rename'
+  end
+end
+
--- a/lib/generators/declare_schema/migration/migrator.rb
+++ b/lib/generators/declare_schema/migration/migrator.rb
@@ -271,7 +271,7 @@ module Generators
           flattened_up = up_commands.flatten.map(&:up)
           up = flattened_up.select(&:present?)
 
-          flattened_down = down_commands.flatten.map(&:down)
+          flattened_down = up_commands.flatten.reverse.map(&:down)
           down = flattened_down.select(&:present?)
 
           [up * "\n\n", down * "\n\n"]
--- a/lib/generators/declare_schema/migration/migrator.rb
+++ b/lib/generators/declare_schema/migration/migrator.rb
@@ -262,7 +262,7 @@ module Generators
           ].compact
 
           <<~EOS.strip
-            create_table :#{model.table_name}, #{create_table_options(model, disable_auto_increment)} do |t|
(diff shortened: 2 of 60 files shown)


lib/generators/declare_schema/migration/migrator.rb
```diff
@@ -201,85 +207,103 @@ def generate
...
+                                ColumnAdd
+                                ColumnRename
+                                ColumnChange
+                                  PrimaryKeyChange
+                                  IndexAdd
+                                    ForeignKeyAdd
+                                    ForeignKeyRemove
+                                  IndexRemove
+                                ColumnRemove
+                              TableRemove ]
```
ColinDKelley, line 280: This is the heart of the ticket. I used indentation here to show dependencies.

spec/lib/declare_schema/migration_generator_spec.rb
```diff
@@ -10,16 +10,6 @@
...
-
-  let(:charset_alter_table) do
-    if defined?(Mysql2)
-      <<~EOS
-
-
-        execute "ALTER TABLE `adverts` CHARACTER SET utf8mb4 COLLATE utf8mb4_bin"
-      EOS
-    end
-  end
```
ColinDKelley, line 22: @jebentier While I was here, I put the CHARACTER SET and COLLATE directly in the `CREATE TABLE` command. The ALTER is now only used when you're changing it.

spec/lib/declare_schema/migration_generator_spec.rb
```diff
@@ -30,6 +20,11 @@
       ', charset: "utf8mb4", collation: "utf8mb4_bin"'
     end
   end
+  let(:create_table_charset_and_collation) do
+    if defined?(Mysql2)
+      ", options: \"CHARACTER SET utf8mb4 COLLATE utf8mb4_bin\""
+    end
+  end
```
ColinDKelley, line 27: Here it is, under `options: `.

spec/lib/declare_schema/migration_generator_spec.rb
```diff
@@ -820,15 +767,8 @@ class Ad < ActiveRecord::Base
...
         rename_table :adverts, :ads
-
         add_column :ads, :created_at, :datetime, null: false
         change_column :ads, :title, :string, limit: 250, null: false, default: \"Untitled\"#{charset_and_collation}
-
-        #{if defined?(SQLite3)
-            "add_index :ads, [:id], unique: true, name: 'PRIMARY'"
-          elsif defined?(Mysql2)
-            'execute "ALTER TABLE ads DROP PRIMARY KEY, ADD PRIMARY KEY (id)"'
-          end}
```
ColinDKelley, line 831: These deleted sections were bugs in the primary key logic.

spec/support/acceptance_spec_helpers.rb
```diff
@@ -45,13 +45,13 @@ def migrate_down(expected_value)
...
     def matches?(subject)
-      super(subject[0].gsub(/, +([a-z_]+:)/i, ', \1')) # normalize multiple spaces to one
+      super(subject[0].strip.gsub(/, +([a-z_]+:)/i, ', \1').gsub(/\n+/, "\n")) # normalize multiple spaces and newlines to one
     end
   end
 
   class MigrationDownEquals < RSpec::Matchers::BuiltIn::Eq
     def matches?(subject)
-      super(subject[1].gsub(/, +([a-z_]+:)/i, ', \1')) # normalize multiple spaces to one
+      super(subject[1].strip.gsub(/, +([a-z_]+:)/i, ', \1').gsub(/\n+/, "\n")) # normalize multiple spaces and newlines to one
```
ColinDKelley, line 54: I made these changes so tests wouldn't fail based on whitespace. I didn't want to waste time on that. We can always tune whitespace some other time.

yinonrousso (approved):
Everything looks great as far as I can tell! These change are much cleaner and lay out a solid dependency structure 👍

ColinDKelley:
@jebentier Would you like to review this as well? I'm ok if you're too busy!

lib/declare_schema/schema_change/all.rb
```diff
@@ -0,0 +1,22 @@
...
+    autoload :ColumnRename, 'declare_schema/schema_change/column_rename'
+    autoload :ForeignKeyAdd, 'declare_schema/schema_change/foreign_key_add'
+    autoload :ForeignKeyRemove, 'declare_schema/schema_change/foreign_key_remove'
+    autoload :IndexAdd, 'declare_schema/schema_change/index_add'
+    autoload :IndexRemove, 'declare_schema/schema_change/index_remove'
+    autoload :PrimaryKeyChange, 'declare_schema/schema_change/primary_key_change'
+    autoload :TableAdd, 'declare_schema/schema_change/table_add'
+    autoload :TableChange, 'declare_schema/schema_change/table_change'
+    autoload :TableRemove, 'declare_schema/schema_change/table_remove'
+    autoload :TableRename, 'declare_schema/schema_change/table_rename'
```
jebentier, line 19: Loving how clean this looks with the `autoload`. I want to start using this pattern more where we can.

jebentier (approved):
Only did a cursory scan of test coverage and structure. I really like how clean all the `SchemaChange` POROs turned out! This is really clean, extensible, and easy to reason about 🎉

lib/declare_schema/schema_change/all.rb
ColinDKelley: Pretty slick, huh? I occasionally see Rails doing it this way.

ColinDKelley:
@jebentier 

> Only did a cursory scan of test coverage and structure.

Not sure you noticed then, but it was cool to use inheritance to extract commonality of the matched pairs like `*Add` and `*Remove` so they share the same code but just reverse the `up` and `down` steps.

PR #69, opened 2021-03-19
Tech 5413 fix bug where constraint is not globally unique

yinonrousso:
### Fixed
- Fixed bug where foreign key constraint names are not globally unique

## TODO
Fix failing `rails-4-mysql` [test](https://github.com/Invoca/declare_schema/runs/2153066804?check_suite_focus=true): 
```
Mysql2::Error: Cannot add foreign key constraint: ALTER TABLE `advertisers` ADD CONSTRAINT `index_advertisers_on_category_id`
       FOREIGN KEY (`category_id`)
         REFERENCES `categories` (`id`)
```

Failure was due to rails 4 [create_table](https://apidock.com/rails/v4.2.1/ActiveRecord/ConnectionAdapters/SchemaStatements/create_table) not recognizing `bigint` type as primary key. With the move to only support rails 5+ we decided to only run this test on rails >= 5

### DIFF
--- a/spec/lib/declare_schema/migration_generator_spec.rb
+++ b/spec/lib/declare_schema/migration_generator_spec.rb
@@ -855,6 +855,7 @@ RSpec.describe 'DeclareSchema Migration Generator' do
         belongs_to :category, limit: 8
       end
     end
+
     it 'will genereate unique constraint names' do
       expect(Generators::DeclareSchema::Migration::Migrator.run).to(
         migrate_up(<<~EOS.strip)
@@ -871,8 +872,8 @@ RSpec.describe 'DeclareSchema Migration Generator' do
           end
           add_index :advertisers, [:category_id], name: :on_category_id
           add_index :affiliates, [:category_id], name: :on_category_id
-          #{"add_foreign_key :advertisers, :categories, column: :category_id, name: :index_advertisers_on_category_id"}
-          #{"add_foreign_key :affiliates, :categories, column: :category_id, name: :index_affiliates_on_category_id"}
+          add_foreign_key :advertisers, :categories, column: :category_id, name: :index_advertisers_on_category_id
+          add_foreign_key :affiliates, :categories, column: :category_id, name: :index_affiliates_on_category_id
           EOS
       )
       migrate
@@ -880,7 +881,7 @@ RSpec.describe 'DeclareSchema Migration Generator' do
       nuke_model_class(Advertiser)
       nuke_model_class(Affiliate)
     end
-  end if defined?(Mysql2)
+  end if !defined?(Sqlite3)
 
   describe 'serialize' do
     before do
--- a/spec/lib/declare_schema/migration_generator_spec.rb
+++ b/spec/lib/declare_schema/migration_generator_spec.rb
@@ -365,10 +365,10 @@ RSpec.describe 'DeclareSchema Migration Generator' do
       migrate_up(<<~EOS.strip)
         add_column :adverts, :category_id, :integer, limit: 8, null: false
         add_index :adverts, [:category_id], name: :on_category_id
-        #{"add_foreign_key :adverts, :categories, column: :category_id, name: :on_category_id" if defined?(Mysql2)}
+        #{"add_foreign_key :adverts, :categories, column: :category_id, name: :index_adverts_on_category_id" if defined?(Mysql2)}
       EOS
       .and migrate_down(<<~EOS.strip)
-        #{"remove_foreign_key :adverts, name: :on_category_id" if defined?(Mysql2)}
+        #{"remove_foreign_key :adverts, name: :index_adverts_on_category_id" if defined?(Mysql2)}
         remove_index :adverts, name: :on_category_id
         remove_column :adverts, :category_id
       EOS
@@ -389,8 +389,8 @@ RSpec.describe 'DeclareSchema Migration Generator' do
       migrate_up(<<~EOS.strip)
         add_column :adverts, :c_id, :integer, limit: 8, null: false
         add_index :adverts, [:c_id], name: :on_c_id
-        #{"add_foreign_key :adverts, :categories, column: :category_id, name: :on_category_id\n" +
-          "add_foreign_key :adverts, :categories, column: :c_id, name: :on_c_id" if defined?(Mysql2)}
+        #{"add_foreign_key :adverts, :categories, column: :category_id, name: :index_adverts_on_category_id\n" +
+          "add_foreign_key :adverts, :categories, column: :c_id, name: :index_adverts_on_c_id" if defined?(Mysql2)}
       EOS
     )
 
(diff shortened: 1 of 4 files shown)


spec/lib/declare_schema/migration_generator_spec.rb
```diff
@@ -835,6 +835,53 @@ class Ad < ActiveRecord::Base
...
+        end
+        belongs_to :category, limit: 8
+      end
+      class Affiliate < ActiveRecord::Base
+        fields do
+          name :string, limit: 250, null: true
+        end
+        belongs_to :category, limit: 8
+      end
+    end
```
ColinDKelley, line 857: I'd throw a blank line here.

spec/lib/declare_schema/migration_generator_spec.rb
```diff
@@ -835,6 +835,53 @@ class Ad < ActiveRecord::Base
...
+            t.integer :category_id, limit: 8, null: false
+          end
+          create_table :affiliates, id: :bigint, options: "CHARACTER SET utf8mb4 COLLATE utf8mb4_bin" do |t|
+            t.string  :name, limit: 250, null: true, charset: "utf8mb4", collation: "utf8mb4_bin"
+            t.integer :category_id, limit: 8, null: false
+          end
+          add_index :advertisers, [:category_id], name: :on_category_id
+          add_index :affiliates, [:category_id], name: :on_category_id
+          #{"add_foreign_key :advertisers, :categories, column: :category_id, name: :index_advertisers_on_category_id"}
+          #{"add_foreign_key :affiliates, :categories, column: :category_id, name: :index_affiliates_on_category_id"}
```
ColinDKelley, line 875: These should have an `if defined?(Sqlite3)` after, right?

ColinDKelley: Or maybe not, since you have an if below. So is the interpolation needed here?

spec/lib/declare_schema/migration_generator_spec.rb
```diff
@@ -835,6 +835,53 @@ class Ad < ActiveRecord::Base
...
+          EOS
+      )
+      migrate
+
+      nuke_model_class(Advertiser)
+      nuke_model_class(Affiliate)
+    end
+  end if defined?(Mysql2)
```
ColinDKelley, line 883: I think elsewhere we use `!defined?(Sqlite3)` for when to apply FKs, don't we? That's based on the assumption that Postgres will be next. It supports FKs.

spec/lib/declare_schema/model/foreign_key_definition_spec.rb
```diff
@@ -25,7 +25,7 @@ class Network < ActiveRecord::Base
     subject { described_class.new(model, foreign_key, options)}
 
     before do
-      allow(connection).to receive(:index_name).with('models', column: 'network_id') { 'on_network_id' }
+      allow(model.connection).to receive(:index_name).with(model.table_name, column: options[:foreign_key]&.to_s || foreign_key.to_s) { 'index_on_network_id' }
```
ColinDKelley, line 28: Is this `with` important for the test? If not, you could just use `with(any_args)` .

spec/lib/declare_schema/model/foreign_key_definition_spec.rb
```diff
@@ -60,6 +60,14 @@ class Network < ActiveRecord::Base
           expect(subject.on_delete_cascade).to be_truthy
         end
       end
+
+      context 'when constraint name passed' do
+        let(:options) { { parent_table: :networks, foreign_key: :the_network_id, index_name: :index_on_network_id,
+                          constraint_name: "", dependent: :delete } }
+        it 'in as ""' do
```
ColinDKelley, line 67: typo?

spec/lib/declare_schema/model/foreign_key_definition_spec.rb
```diff
@@ -60,6 +60,14 @@ class Network < ActiveRecord::Base
           expect(subject.on_delete_cascade).to be_truthy
         end
       end
+
+      context 'when constraint name passed' do
+        let(:options) { { parent_table: :networks, foreign_key: :the_network_id, index_name: :index_on_network_id,
```
ColinDKelley, line 65: I thought we were getting rid of the `index_name:` option? I think we only need `constraint_name:`.

spec/lib/declare_schema/model/foreign_key_definition_spec.rb
yinonrousso: Great point!

spec/lib/declare_schema/model/foreign_key_definition_spec.rb
yinonrousso: Great point, there's no need for the index_name option anymore!

spec/lib/declare_schema/migration_generator_spec.rb
yinonrousso: Hmm I'm getting some unexpected behavior where `Sqlite3` is always `nil`
Screen Shot 2021-03-19 at 1 02 50 PM

ColinDKelley: Ah, that's because of their weird capitalization where they matched their branding. It's `SQLite3`:

https://github.com/Invoca/declare_schema/blob/df444042a5d5b6d7720067cde66652485c82cd53/spec/lib/declare_schema/generator_spec.rb#L104

ColinDKelley (approved):
Looks good, pending green tests of course!

spec/lib/declare_schema/migration_generator_spec.rb
yinonrousso: 🤦  Ooof. There's a couple of other spots that use the it with a lowercase. I'll go ahead and update those as well

ColinDKelley: Oh, weird! So those places were presumably not running at all, given that casing. Hope they still pass!

ColinDKelley:
commit message: cosntant -> constant

spec/lib/declare_schema/migration_generator_spec.rb
yinonrousso: Interesting the [sqlite appraisal](https://github.com/Invoca/declare_schema/pull/69/checks?check_run_id=2152326380) is still running with the `if !defined?(SQLite3)` condition

ColinDKelley: Oh, I wonder if it hasn't been required at that point? That was a problem with checking mysql and we wound up with this hack in a few places:
```
begin
  require 'mysql2'
  require 'active_record/connection_adapters/mysql2_adapter'
rescue LoadError
end
```

ColinDKelley: There's also the approach of asking the driver directly (which implicitly requires the driver gems):
```
if ActiveRecord::Base.connection.class.name.match?(/SQLite3Adapter/)
```
This is the way we have to do it for actual running code (not tests) since Rails allows multiple drivers to be configured (and hence all their gems would be loaded).

yinonrousso: That makes a great deal of sense. Looks like it indeed had not been required at that point. Curious as to why we can't use the same driver approach in tests? I get a `ActiveRecord::ConnectionNotEstablished: No connection pool with 'primary' found.` when I try using in it in this test. Does the dummy test app just not have a connection established at this point?

ColinDKelley: I don't fully understand when these tests are able to connect to the database. It's a very unusual setup compared to normal `rspec`. So far I've just worked around it.

PR #70, opened 2021-03-19
no-jira:drop generate and migrate command

ColinDKelley:
### Removed
- Removed `g|m|c` prompt entirely, since it was confusing. Instead, the migration is always generated; the user may press ^C at the filename prompt to cancel. The migration will be run if `--migrate` is passed; otherwise, the migrate command will be displayed to be run later.
### Added
- Added the new configuration option `DeclareSchema.@db_migrate_command =`.

### DIFF
--- a/lib/declare_schema/model.rb
+++ b/lib/declare_schema/model.rb
@@ -1,7 +1,5 @@
 # frozen_string_literal: true
 
-require 'rails'
-
 require 'declare_schema/extensions/module'
 
 module DeclareSchema
@@ -130,7 +128,7 @@ module DeclareSchema
 
         fk_options[:dependent] = options.delete(:far_end_dependent) if options.has_key?(:far_end_dependent)
 
-        if Rails::VERSION::MAJOR >= 5
+        if ActiveSupport::VERSION::MAJOR >= 5
           super
         else
           super(name, scope, options.except(:optional))
@@ -149,7 +147,7 @@ module DeclareSchema
         end
       end
 
-      if ::Rails::VERSION::MAJOR < 5
+      if ::ActiveSupport::VERSION::MAJOR < 5
         def primary_key
           super || 'id'
         end
@@ -227,7 +225,7 @@ module DeclareSchema
                   ActiveRecord::Coders::JSON
                 elsif [:load, :dump].all? { |x| class_name_or_coder.respond_to?(x) }
                   class_name_or_coder
-                elsif Rails::VERSION::MAJOR >= 5
+                elsif ActiveSupport::VERSION::MAJOR >= 5
                   ActiveRecord::Coders::YAMLColumn.new(attr_name, class_name_or_coder)
                 else
                   ActiveRecord::Coders::YAMLColumn.new(class_name_or_coder)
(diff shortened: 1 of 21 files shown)


lib/generators/declare_schema/migration/migration_generator.rb
```diff
@@ -67,51 +65,55 @@ def migrate
...
+
+      @up = indent(up, 4)
+      @down = indent(down, 4)
+      @migration_class_name = final_migration_name.camelize
+
+      migration_template('migration.rb.erb', "db/migrate/#{final_migration_name.underscore}.rb")
+
+      db_migrate_command = ::DeclareSchema.db_migrate_command
+      if options[:migrate]
+        say db_migrate_command
```
dcaddell, line 84: `say`ing the migration command before we strip the `bundle exec rails|rake` seems like it might cause some confusion down the line. If we want this sort of logging, what about moving it to the `rake` and `rails_command` methods?

dcaddell (approved):
One comment about some logging but otherwise LGTM

lib/declare_schema/model.rb
```diff
@@ -130,7 +128,7 @@ def belongs_to(name, scope = nil, **options)
 
         fk_options[:dependent] = options.delete(:far_end_dependent) if options.has_key?(:far_end_dependent)
 
-        if Rails::VERSION::MAJOR >= 5
+        if ActiveSupport::VERSION::MAJOR >= 5
```
yinonrousso, line 131: Am I understanding correctly that you're cleaning up the explicit need to `require rails` and falling back to the autoloaded `activesupport` instead?

yinonrousso (approved):
Changes look great!

lib/generators/declare_schema/migration/migration_generator.rb
```diff
@@ -38,20 +38,18 @@ def banner
...
                  aliases: '-m',
                  type: :boolean,
-                 desc: "Don't prompt for action - generate and migrate"
+                 desc: "After generating migration, run it"
 
     def migrate
       return if migrations_pending?
 
-      generator = Generators::DeclareSchema::Migration::Migrator.new(->(c, d, k, p) { extract_renames!(c, d, k, p) })
+      generator = Generators::DeclareSchema::Migration::Migrator.new do |to_create, to_drop, kind_str, name_prefix|
```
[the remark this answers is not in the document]
dcaddell: Yeah this change is the stealthy hero of this PR imo

lib/generators/declare_schema/migration/migration_generator.rb
ColinDKelley: My reasoning is that you need the `bundle exec rails|rake` if you run from the command line...but internally, that prefix has already been run so we don't need to (aren't allowed to, in fact) repeat it.

lib/generators/declare_schema/migration/migration_generator.rb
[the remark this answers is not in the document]
ColinDKelley: Yeah, I don't know what they were thinking abbreviating those to one letter. And why were they passing a block as a parameter when Ruby has a built-in `&block` parameter that already does exactly that?

lib/declare_schema/model.rb
ColinDKelley: Yes, the tests failed on CI because rails wasn't explicitly required at this point. I glanced and found that ActiveSupport exposes the same version, so I figured we could just depend on that. Currently this entire gem still depends on `rails` because of `generators`, but not all the pieces need to.

PR #71, opened 2021-03-26
no-jira: fix empty up/down bug

ColinDKelley:
### Fixed
- Fixed a bug where up and down in generated migration would be empty in Rails 4.

### DIFF
--- a/lib/generators/declare_schema/migration/migration_generator.rb
+++ b/lib/generators/declare_schema/migration/migration_generator.rb
@@ -73,8 +73,8 @@ module DeclareSchema
         end ||
         default_migration_name
 
-      @up = indent(up, 4)
-      @down = indent(down, 4)
+      @up = indent(up.strip, 4)
+      @down = indent(down.strip, 4)
       @migration_class_name = final_migration_name.camelize
 
       migration_template('migration.rb.erb', "db/migrate/#{final_migration_name.underscore}.rb")
@@ -100,7 +100,7 @@ module DeclareSchema
     if ActiveSupport::VERSION::MAJOR < 5
       def indent(string, columns)
         whitespace = ' ' * columns
-        string.gsub("\n", "\n#{whitespace}").gsub!(/ +\n/, "\n")
+        string.gsub("\n", "\n#{whitespace}").gsub(/ +\n/, "\n")
       end
     end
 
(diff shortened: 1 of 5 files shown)


lib/generators/declare_schema/migration/migration_generator.rb
```diff
@@ -73,8 +73,8 @@ def migrate
         end ||
         default_migration_name
 
-      @up = indent(up, 4)
-      @down = indent(down, 4)
+      @up = indent(up.strip, 4)
+      @down = indent(down.strip, 4)
```
ColinDKelley, line 77: While I was here, I decided to strip the empty lines at the end of the `up` and `down` methods since they looked a bit distracting, and hound would have complained about them.

lib/generators/declare_schema/migration/migration_generator.rb
```diff
@@ -100,7 +100,7 @@ def migrate
     if ActiveSupport::VERSION::MAJOR < 5
       def indent(string, columns)
         whitespace = ' ' * columns
-        string.gsub("\n", "\n#{whitespace}").gsub!(/ +\n/, "\n")
+        string.gsub("\n", "\n#{whitespace}").gsub(/ +\n/, "\n")
```
ColinDKelley, line 103: Here's the bug. One `!`. At the last minute I changed to `gsub!` as a tiny optimization, since the string has already been dup'd by the previous `gsub`. Bad idea! In the event that the pattern isn't found, `gsub!` returns `nil` instead of the original string. I had no idea of that weird contract.

This would only fail in Rails 4, and only on migrations that were a single line.

christinalavingia: Haha glad my use case was the perfect example to spot this bug!

christinalavingia (approved):
Seems like a straightforward fix, thank you for pushing this Colin!

lib/generators/declare_schema/migration/migration_generator.rb
dcaddell: Didn't understand what you meant by "weird contract" so I ran it through irb, and yeah, this is sort of infuriating. Good spot.
```
irb(main):001:0> "pizza".gsub("party", "123")
=> "pizza"
irb(main):002:0> "pizza".gsub!("party", "123")
=> nil
```

dcaddell (approved):

lib/generators/declare_schema/migration/migration_generator.rb
ColinDKelley: Yeah, that was a bizarre contract choice. I checked and it is documented, though. I guess the reason I didn't know is that I almost always avoid the `!` methods.

PR #76, opened 2021-04-01
Update README to use `declare_schema` instead of `fields`

nburwell:
I noticed the readme was still out of date with the deprecated `fields` DSL


dcaddell (approved):
Thank you! 🙌

ColinDKelley (approved):

PR #81, opened 2021-04-27
TECH-5417: infer foreign key limit

ColinDKelley:
Note: best to view with `Hide whitespace changes` since there are some indentation changed in tests.

https://ringrevenue.atlassian.net/browse/TECH-5417

Infers foreign key size from the corresponding primary key. In the case of polymorphic keys, this isn't possible, so it always uses `limit: 8` there, unless the schema has already been migrated with `limit: 4`.
```
        # Note: the foreign key limit: should match the primary key limit:. (If there is a foreign key constraint,
        # those limits _must_ match.) We'd like to call _infer_fk_limit and get the limit right from the PK.
        # But we can't here, because that will mess up the autoloader to follow every belongs_to association right
        # when it is declared. So instead we assume :bigint (integer limit: 8) below, while also registering this
        # pre_migration: callback to double-check that assumption Just In Time--right before we generate a migration.
        #
        # The one downside of this approach is that application code that asks the field_spec for the declared
        # foreign key limit: will always get 8 back even if this is a grandfathered foreign key that points to
        # a limit: 4 primary key. It seems unlikely that any application code would do this.
```

### DIFF
--- a/lib/declare_schema/extensions/active_record/fields_declaration.rb
+++ b/lib/declare_schema/extensions/active_record/fields_declaration.rb
@@ -7,12 +7,14 @@ require 'declare_schema/field_declaration_dsl'
 
 module DeclareSchema
   module Macros
+    attr_reader :_table_options
+
     def fields(table_options = {}, &block)
       # Any model that calls 'fields' gets DeclareSchema::Model behavior
       DeclareSchema::Model.mix_in(self)
 
       @include_in_migration = true
-      @table_options        = table_options
+      @_table_options        = table_options
 
       if block
         dsl = DeclareSchema::FieldDeclarationDsl.new(self)
@@ -31,7 +33,7 @@ module DeclareSchema
 
       # @include_in_migration = false #||= options.fetch(:include_in_migration, true); options.delete(:include_in_migration)
       @include_in_migration = true # TODO: Add back or delete the include_in_migration feature
-      @table_options        = table_options
+      @_table_options        = table_options
 
       if block
         dsl = DeclareSchema::Dsl.new(self, null: false)
--- a/lib/declare_schema/model.rb
+++ b/lib/declare_schema/model.rb
@@ -183,7 +183,7 @@ module DeclareSchema
               4
             end
           else
-            if klass.table_exists? && (pk_column = klass.columns_hash[klass._defined_primary_key])
+            if klass.table_exists? && (pk_column = klass.columns_hash[klass._declared_primary_key])
               pk_id_type = pk_column.type
               if pk_id_type == :integer
                 pk_column.limit
@@ -201,27 +201,27 @@ module DeclareSchema
 
       # returns the primary key (String) as declared with primary_key =
       # unlike the `primary_key` method, DOES NOT query the database to find the actual primary key in use right now
-      # if no explicit primary key set, returns the default_defined_primary_key
-      def _defined_primary_key
+      # if no explicit primary key set, returns the _default_declared_primary_key
+      def _declared_primary_key
         if defined?(@primary_key)
           @primary_key&.to_s
-        end || _default_defined_primary_key
+        end || _default_declared_primary_key
       end
 
       private
 
-      # if this is a derived class, returns the base class's _defined_primary_key
+      # if this is a derived class, returns the base class's _declared_primary_key
       # otherwise, returns 'id'
-      def _default_defined_primary_key
+      def _default_declared_primary_key
         if self == base_class
           'id'
         else
-          base_class._defined_primary_key
+          base_class._declared_primary_key
         end
       end
 
       def _rails_default_primary_key
-        ::DeclareSchema::Model::IndexDefinition.new(self, [_defined_primary_key.to_sym], unique: true, name: DeclareSchema::Model::IndexDefinition::PRIMARY_KEY_NAME)
+        ::DeclareSchema::Model::IndexDefinition.new(self, [_declared_primary_key.to_sym], unique: true, name: DeclareSchema::Model::IndexDefinition::PRIMARY_KEY_NAME)
       end
 
       # Declares the "foo_type" field that accompanies the "foo_id"
--- a/lib/declare_schema/model.rb
+++ b/lib/declare_schema/model.rb
@@ -142,17 +142,24 @@ module DeclareSchema
 
         refl = reflections[name.to_s] or raise "Couldn't find reflection #{name} in #{reflections.keys}"
         fkey = refl.foreign_key or raise "Couldn't find foreign_key for #{name} in #{refl.inspect}"
-        klass = refl.klass or raise "Couldn't find belongs_to klass for #{name} in #{refl.inspect}"
-        unless (pk_id_type = klass._table_options&.[](:id))
-          if klass.table_exists? && (pk_column = klass.columns_hash[klass._defined_primary_key])
-            pk_id_type = pk_column.type
-            if pk_id_type == :integer
-              column_options[:limit] = pk_column.limit
-            end
-          end
+        column_options[:pre_migration] = ->(field_spec) do
+          klass = refl.klass or raise "Couldn't find belongs_to klass for #{name} in #{refl.inspect}"
+          new_limit =
+            if (pk_id_type = klass._table_options&.[](:id))
+              if pk_id_type == :integer
+                4
+              end
+            else
+              if klass.table_exists? && (pk_column = klass.columns_hash[klass._defined_primary_key])
+                pk_id_type = pk_column.type
+                if pk_id_type == :integer
+                  pk_column.limit
+                end
+              end
+            end and field_spec.sql_options[:limit] = new_limit
         end
 
-        declare_field(fkey.to_sym,  pk_id_type || :bigint, column_options)
+        declare_field(fkey.to_sym, :bigint, column_options)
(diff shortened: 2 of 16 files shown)


lib/declare_schema/extensions/active_record/fields_declaration.rb
```diff
@@ -7,12 +7,14 @@
 
 module DeclareSchema
   module Macros
+    attr_reader :_table_options
```
ColinDKelley, line 10: I realized we were clashing with Rails own `table_options`. No need for that.

lib/declare_schema/model.rb
```diff
@@ -155,27 +201,27 @@ def primary_key
 
       # returns the primary key (String) as declared with primary_key =
       # unlike the `primary_key` method, DOES NOT query the database to find the actual primary key in use right now
-      # if no explicit primary key set, returns the default_defined_primary_key
-      def _defined_primary_key
+      # if no explicit primary key set, returns the _default_declared_primary_key
+      def _declared_primary_key
```
ColinDKelley, line 205: I realized "declared" is a better adjective, given this gem name.

lib/declare_schema/model/column.rb
```diff
@@ -59,7 +59,7 @@ def deserialize_default_value(column, type, default_value)
             # might be getting migrated to a new type. We should be using just type as below. -Colin
             column.type_cast_from_database(default_value)
           else
-            cast_type = ActiveRecord::Base.connection.send(:lookup_cast_type, type) or
+            cast_type = ActiveRecord::Base.connection.send(:lookup_cast_type, type.to_s) or
```
ColinDKelley, line 62: The Rails 5 migration rider.

lib/declare_schema/model/index_definition.rb
```diff
@@ -67,7 +67,7 @@ def index_name(columns)
         private
 
         # This is the old approach which is still needed for MySQL in Rails 4 and SQLite
-        def sqlite_compound_primary_key(model, table)
+        def fallback_find_primary_key(model, table)
```
ColinDKelley, line 70: Renamed since it also covers MySQL.

lib/declare_schema/model/index_definition.rb
```diff
@@ -83,9 +83,11 @@ def each_hash(result)
             end
           end
 
-          pk_index = connection.indexes(table).find { |index| index.name.to_s == PRIMARY_KEY_NAME } or return nil
-
-          Array(pk_index.columns)
+          if (pk_index = connection.indexes(table).find { |index| index.name.to_s == PRIMARY_KEY_NAME })
+            Array(pk_index.columns)
+          elsif model.connection.columns(table).any? { |col| col.name == 'id' }
+            ['id']
```
ColinDKelley, line 89: This is the right answer 99.999% of the time, so we might as well fall back to it.

lib/generators/declare_schema/migration/migrator.rb
```diff
@@ -187,6 +187,12 @@ def generate
           models, db_tables = models_and_tables
           models_by_table_name = {}
           models.each do |m|
+            m.try(:field_specs)&.each do |_name, field_spec|
+              if (pre_migration = field_spec.options.delete(:pre_migration))
+                pre_migration.call(field_spec)
```
ColinDKelley, line 192: This is the callback hook that lets the field spec update the `limit:` to 4 if it needs to. Waiting until this late is necessary to traverse to the other side of the `belongs_to`.

lib/generators/declare_schema/migration/migrator.rb
```diff
@@ -336,18 +342,17 @@ def change_table(model, current_table_name)
           new_table_name = model.table_name
 
           db_columns = model.connection.columns(current_table_name).index_by(&:name)
-          key_missing = db_columns[model._defined_primary_key].nil? && model._defined_primary_key.present?
-          if model._defined_primary_key.present?
-            db_columns.delete(model._defined_primary_key)
+          if (pk = model._declared_primary_key.presence)
+            key_was_in_db_columns = db_columns.delete(pk)
```
ColinDKelley, line 346: I finally untangled this rat's nest. Turns out it was much simpler than it looked.

lib/generators/declare_schema/migration/migrator.rb
```diff
@@ -336,18 +342,17 @@ def change_table(model, current_table_name)
...
 
           model_column_names = model.field_specs.keys.map(&:to_s)
           db_column_names = db_columns.keys.map(&:to_s)
 
           to_add = model_column_names - db_column_names
-          to_add += [model._defined_primary_key] if key_missing && model._defined_primary_key.present?
+          to_add += [pk] unless key_was_in_db_columns
           to_remove = db_column_names - model_column_names
-          to_remove -= [model._defined_primary_key.to_sym] if model._defined_primary_key.present?
+          to_remove -= [pk.to_sym] if pk    # TODO: The .to_sym here means this code is always a no-op, right? -Colin
```
ColinDKelley, line 355: Can you see any way that the `-=` would ever do anything? It seems to me that the `to_remove` array must always be Strings.

spec/lib/declare_schema/api_spec.rb
```diff
@@ -41,10 +41,12 @@ class Advert < #{active_record_base_class}
 
       if ActiveSupport::VERSION::MAJOR == 5
         # TODO: get this to work on Travis for Rails 6
+        # TODO: uncomment since we're not on Travis any more? -Colin
```
ColinDKelley, line 44: Seems like we should try sometime.

spec/lib/declare_schema/migration_generator_spec.rb
```diff
@@ -2039,7 +2039,6 @@ class Affiliate < ActiveRecord::Base
           add_foreign_key :affiliates, :categories, column: :category_id, name: :index_affiliates_on_category_id
         EOS
         )
-        binding.pry
```
ColinDKelley, line 2042: Doh!

lib/declare_schema/model/index_definition.rb
jebentier: Can we ask the model what the `_declared_primary_key` or use Rails' built in methodology for looking up with the primary key for a given model is here? I agree that 99.999% of the time this is correct, but I also don't think we need to go fishing for the primary key index in the database do we?

lib/generators/declare_schema/migration/migrator.rb
jebentier: Yea it does look like the `to_sym` is always making this a no-op. Everything is `to_s`'d before coming to this line. I get that this line is trying to remove the primary key from the list of potential columns to remove. This might be a latent bug that we need to fix by changing this to `to_remove -= [pk]`, because right now as this sits, it's possible that you can remove the primary key.

spec/lib/declare_schema/api_spec.rb
jebentier: Wanna try this just before merge and if it works merge if it doesn't revert and create a new ticket?

jebentier (commented):
Overall looks good, just have a couple implementation questions

lib/generators/declare_schema/migration/migrator.rb
ColinDKelley: ^ @jebentier Do you agree? I could change it to just `pk.to_s`...but apparently no one is missing that it doesn't work that way now... so maybe it never happens that the PK column name is in this list at all?

ColinDKelley: Oh hey, if you scroll up you can see that `to_remove` started from `db_column_names`, which came from `db_columns`, which already had the `pk` removed. Untangling the rat's nest is paying off already. 🎉 
I'll get rid of this line.

lib/declare_schema/model/index_definition.rb
ColinDKelley: Yes, we need to look in the database because that's the _only_ reason this method is there. We have `_declared_primary_key` for when we want to know what was declared.

Want to change any method names to help make this clear?

BTW I'm looking forward to deleting some of this code as soon as we stop supporting Rails 4. This method will live on, but only for `sqlite`.

spec/lib/declare_schema/api_spec.rb
ColinDKelley: Yeah, I can try.

lib/declare_schema/model/index_definition.rb
jebentier: No, don't think any renaming is necessary. What you explained makes sense, a common mixup that's easy to happen throughout this code is differentiating between the cases where you need to access the declared vs the actual. Thank you for the explanation, I don't think any changes are necessary here and look forward to the day of this not being as complicated.

lib/generators/declare_schema/migration/migrator.rb
jebentier: Awesome! 🎉

jebentier (approved):
LGTM!

spec/lib/declare_schema/api_spec.rb
ColinDKelley: It didn't work. And I'm actually not sure why that line is needed there for Rails 5?

So for now I deleted the comments here, and also the one below about rich types.

PR #85, opened 2021-05-09
no-jira: enforce enum symbols

ColinDKelley:
### Fixed
- When an `enum` type field is declared, there is now enforcement that its `limit:` must be an array of 1 or more Symbols,
  and its `default:`--if given--must be a Symbol or `nil`.


alecjacobs5401 (approved):
👏

PR #88, opened 2021-06-08
TECH-5757: default_schema

ColinDKelley:
Passion project: TECH-5757

### Added
- Added support for `default_schema` to apply a default schema to every model, unless disabled for that model with `default_schema: false`.

### DIFF
--- a/lib/declare_schema/dsl.rb
+++ b/lib/declare_schema/dsl.rb
@@ -7,7 +7,7 @@ module DeclareSchema
     include ::Kernel      # but we need the basic class methods
 
     instance_methods.each do |m|
-      unless m.to_s.starts_with?('__') || m.in?([:object_id, :instance_eval])
+      unless m.to_s.starts_with?('__') || m.in?([:object_id, :instance_eval, :instance_exec])
         undef_method(m)
       end
     end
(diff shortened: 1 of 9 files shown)


lib/declare_schema/dsl.rb
```diff
@@ -32,8 +32,11 @@ def field(name, type, *args, **options)
       @model.declare_field(name, type, *(args + [@options.merge(options)]))
     end
 
-    def method_missing(type, name, *args)
-      field(name, type, *args)
+    # TODO: make [:required] just another option. Either 'required: true] or 'optional: false'?
+    def method_missing(*args, **options)
+      args.count(&:itself) >= 2 or raise ::ArgumentError, "fields in declare_schema block must be declared as: type name, [:required], options (got #{args.inspect}, #{options.inspect})"
```
yinonrousso, line 37: NIT: Missing comma in error message
```suggestion
      args.count(&:itself) >= 2 or raise ::ArgumentError, "fields in declare_schema block must be declared as: type, name, [:required], options (got #{args.inspect}, #{options.inspect})"
```

yinonrousso (approved):
LGTM! This is a really clean implementation 👍

lib/declare_schema/dsl.rb
ColinDKelley: Actually there is no comma there because the `type` is the method name. As in
```
integer :percent_allowed
```

PR #97, opened 2021-08-02
TECH-6171: fix indentation and triple spacing

ColinDKelley:
### Fixed
- Fixed migration file indentation bug in Rails 5 where the first line was indented an extra 4 characters.
 And fixed the longstanding triple-spacing bug to use double spacing.

# Demo
## Before
```
class DeclareSchemaMigration1 < (ActiveSupport::VERSION::MAJOR >= 5 ? ActiveRecord::Migration[4.2] : ActiveRecord::Migration)
  def self.up
        add_column :deploy_targets, :a, :string, limit: 10, null: false, charset: "utf8mb4", collation: "utf8mb4_general_ci"


    add_column :deploy_targets, :b, :string, limit: 15, null: false, charset: "utf8mb4", collation: "utf8mb4_general_ci"
  end

  def self.down
        remove_column :deploy_targets, :b


    remove_column :deploy_targets, :a
  end
end

```
## After
```
class DeclareSchemaMigration1 < (ActiveSupport::VERSION::MAJOR >= 5 ? ActiveRecord::Migration[4.2] : ActiveRecord::Migration)
  def self.up
    add_column :deploy_targets, :a, :string, limit: 10, null: false, charset: "utf8mb4", collation: "utf8mb4_general_ci"

    add_column :deploy_targets, :b, :string, limit: 15, null: false, charset: "utf8mb4", collation: "utf8mb4_general_ci"
  end

  def self.down
    remove_column :deploy_targets, :b

    remove_column :deploy_targets, :a
  end
end
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ Inspired by [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 Note: this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.13.1] - Unreleased
+### Fixed
+- Fixed migration file indentation bug in Rails 5 where the first line was indented an extra 4 characters.
+ And fixed the longstanding triple-spacing bug to use double spacing.
+
 ## [0.13.0] - 2020-06-11
 ### Added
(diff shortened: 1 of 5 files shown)


dcaddell (approved):
👏 🥳

PR #98, opened 2021-08-04
TECH-6162 - fix support for disable auto increment

primiti:
TECH-6162 - add call fact shards

This is a fix for an issue that I ran into adding new call fact shards.    The primary key definition was not being added as a nested array. 

Without this fix, creating new warehouse shards fails with the following error.  Note that the first element is not a nested array.

```fields must be Array(Array(Symbol, Symbol, Hash)); got [:integer, :id, {:limit=>8, :auto_increment=>false, :primary_key=>true}, [:datetime, :created_at, {:null=>true}],...```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ Inspired by [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 Note: this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.13.2] - 2021-08-04
+### Fixed
+- Fixed issue with disable_auto_increment support for new tables.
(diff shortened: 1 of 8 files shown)


ColinDKelley (approved):
Looks good!

primiti:
Just a heads up for @ColinDKelley - There were appraisal errors that I did not notice in the code you reviewed.  I iterated on it with a couple commits and now have a green build.  I did not change production code, just the test. I am going to push because I want to get my changes into the mainline, but if you have any feedback I will make the changes on a new branch.  Thanks.

ColinDKelley:
Good call! The changes look fine. Glad you didn't let Rubocop bully you into using `unless`.

PR #99, opened 2021-08-05
TECH-6149 drop support for rails 4

irismoini:


spec/lib/declare_schema/migration_generator_spec.rb
```diff
@@ -2461,13 +2425,13 @@ class Advert < active_record_base_class.constantize
...
         base_class = first_line.split(' < ').last
-        expect(base_class).to eq("(ActiveSupport::VERSION::MAJOR >= 5 ? ActiveRecord::Migration[4.2] : ActiveRecord::Migration)")
+        expect(base_class).to eq("(ActiveRecord::Migration[4.2])")
       end
     end
 
     context 'Does not generate migrations' do
       it 'for aliased fields bigint -> integer limit 8' do
-        if ActiveSupport::VERSION::MAJOR >= 5 || !ActiveRecord::Base.connection.class.name.match?(/SQLite3Adapter/)
+        if !ActiveRecord::Base.connection.class.name.match?(/SQLite3Adapter/)
```
dcaddell, line 2434: I think we should delete this `if` condition since the deleted portion `ActiveSupport::VERSION::MAJOR >= 5 ||` would always evaluate to true.

spec/lib/declare_schema/model/column_spec.rb
```diff
@@ -63,7 +60,6 @@
     end
 
     describe '.deserialize_default_value' do
```
dcaddell, line 62: It's hard to tell for sure, but I think the tests inside this block need to un-tabbed since the `if` was deleted

dcaddell (changes requested):
Really great job on this! I noticed a couple of missing changes but I'm really impressed overall!

- We'll need to update the gem version and make some additions to the CHANGELOG. 
- [This comment](https://github.com/Invoca/declare_schema/blob/6c4331d65bad99cac136b69ac97e8bfee9e7746b/lib/declare_schema/model/index_definition.rb#L69) leads me to believe that this method should be deleted.
- There's two ActiveRecord::VERSION references in the `migration_generator_spec`. [Here](https://github.com/Invoca/declare_schema/blob/35b19280a1d93a3c8c0f64a533304888c385ff13/spec/lib/declare_schema/migration_generator_spec.rb#L879) and [here](https://github.com/Invoca/declare_schema/blob/35b19280a1d93a3c8c0f64a533304888c385ff13/spec/lib/declare_schema/migration_generator_spec.rb#L2018).

CHANGELOG.md
```diff
@@ -3,6 +3,9 @@
 Inspired by [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 Note: this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+## [0.14.0] - 2021-08-09
```
dcaddell, line 6: We'll need to update this date when we actually release the change (probably tomorrow). We would generally leave the date here as `Unreleased` until then but it's not a huge deal.

Also, if you check the bottom of this file you'll see a bunch of lines that look like `[version_number]: link_to_github_diff`. Mind adding one for this change entry? That's the markdown that'll turn the [0.14.0] into a hyperlink.

CHANGELOG.md
```diff
@@ -3,6 +3,9 @@
...
+### Deprecated
+- Support for Rails < 5
```
dcaddell, line 8: Full sentences are preferred for CHANGELOG entries.

dcaddell (approved):
Nice job, just a couple nitpicks about the changelog entries

dcaddell (approved):
🎉

PR #106, opened 2021-09-01
non-production/use ruby 2.7 and fix deprecations

ColinDKelley:
### Fixed
- Fixed Ruby 2.7 warning about needing `**options` when calling a method that has `**options` in its signature.


irismoini (approved):
LGTM 🤠

PR #130, opened 2022-03-23
TECH-7284: update ruby and bundler versions

jebentier:


jebentier:
I'm still working on the 3.x support, it's taking some time so wanted to get this open and going first so that it's easier to cut bait later if necessary

lib/declare_schema/extensions/active_record/fields_declaration.rb
```diff
@@ -9,24 +9,6 @@ module DeclareSchema
...
-        dsl = DeclareSchema::FieldDeclarationDsl.new(self)
-        if block.arity == 1
-          yield dsl
-        else
-          dsl.instance_eval(&block)
-        end
-      end
-    end
-    deprecate :fields, deprecator: ActiveSupport::Deprecation.new('1.0', 'DeclareSchema')
-
```
ColinDKelley, line 29: Nice that you dropped this! 💯

ColinDKelley (approved):
Really nice cleanup! That's cool that `**` worked so cleanly everywhere and didn't break any code on older versions of Ruby. Now we're set up in future PRs to actually start nailing down the options as actual keywords...when they're finite...which I think is most of the time.

PR #148, opened 2022-07-17
non-production: autoload column_remove in schem_change all

jebentier:
### Fixed
- Fixed a bug where `SchemaChange::ColumnRemove` was not loaded and causing an exception to raise
when a column was being removed during migration generation


ColinDKelley (approved):
Oops! Looks good.

ColinDKelley:
@jebentier While you're here, can you remove dependabot config?

PR #149, opened 2022-07-19
TECH-7766: [SPIKE] get change working for deploybot branch; go back a…

yinonrousso:
## Description 
This work came up as part of adding the habtm model relation between `Commits` and `Branches` in [TECH-7766](https://invoca.atlassian.net/browse/TECH-7766). Fix the generator for this case and all habtm relations.


lib/declare_schema/model/habtm_model_shim.rb
```diff
@@ -66,8 +69,8 @@ def ignore_indexes
 
       def constraint_specs
         [
-          ForeignKeyDefinition.new(self, foreign_keys.first, parent_table: foreign_key_classes.first.table_name, constraint_name: "#{join_table}_FK1", dependent: :delete),
-          ForeignKeyDefinition.new(self, foreign_keys.last, parent_table: foreign_key_classes.last.table_name, constraint_name: "#{join_table}_FK2", dependent: :delete)
+          ForeignKeyDefinition.new(self, foreign_keys.first, parent_table: foreign_key_classes.first.table_name, constraint_name: "#{join_table}_FK1"),
+          ForeignKeyDefinition.new(self, foreign_keys.last, parent_table: foreign_key_classes.last.table_name, constraint_name: "#{join_table}_FK2")
```
yinonrousso, line 73: I'm under the impression that the we do **not** want to delete the associations because they can belong to other objects

lib/declare_schema/model/index_definition.rb
```diff
@@ -33,28 +33,30 @@ def initialize(model, fields, **options)
...
         # extract IndexSpecs from an existing table
-        # always includes the PRIMARY KEY index
+        # always includes the PRIMARY KEY index for Model unless it is a HABTM Model.
         def for_model(model, old_table_name = nil)
           t = old_table_name || model.table_name
 
-          primary_key_columns = Array(model.connection.primary_key(t)).presence or
+          habtm_model = model.is_a?(DeclareSchema::Model::HabtmModelShim)
+
+          primary_key_columns = Array(model.connection.primary_key(t)).presence || habtm_model or
```
yinonrousso, line 42: It's a bit crude and uninformative to have the `primary_key_columns` be set to `true` when the `habtm_model` exists. TODO: make this better

lib/declare_schema/model/habtm_model_shim.rb
```diff
@@ -53,8 +57,7 @@ def _declared_primary_key
 
       def index_definitions_with_primary_key
         [
-          IndexDefinition.new(self, foreign_keys, unique: true, name: ::DeclareSchema::Model::IndexDefinition::PRIMARY_KEY_NAME),
-          IndexDefinition.new(self, foreign_keys.last) # not unique by itself; combines with primary key to be unique
+          IndexDefinition.new(self, foreign_keys, unique: true, name: index_name) # creates a primary composite key on both foriegn keys
```
yinonrousso, line 60: `index_definitions_with_primary_key` is a method used on both Model and HabtmModel. This may now read better with a new singular name like `index_definition` but also I don't want to stray too far since the migrator generator uses this accessor for both classes.

lib/declare_schema/model/habtm_model_shim.rb
jebentier: I think we do still want the dependent delete put in. That way if one of the linked objects is deleted, we don't have the join record hanging around pointing to nothing.

lib/declare_schema/model/habtm_model_shim.rb
jebentier: Yea, lets leave the naming as is for now and we can clean that up later if we so choose.

lib/declare_schema/model/index_definition.rb
jebentier: we can break this into two lines to make it cleaner
```suggestion
          primary_key_columns = Array(model.connection.primary_key(t)).presence 
          primary_key_columns || habtm_model or raise ...
```

lib/declare_schema/model/index_definition.rb
```diff
@@ -33,28 +33,30 @@ def initialize(model, fields, **options)
...
             raise "could not find primary key for table #{t} in #{model.connection.columns(t).inspect}"
 
           primary_key_found = false
           index_definitions = model.connection.indexes(t).map do |i|
             model.ignore_indexes.include?(i.name) and next
-            if i.name == PRIMARY_KEY_NAME
+            if i.name == PRIMARY_KEY_NAME && !habtm_model
```
jebentier, line 48: will the habtm_model ever return a primary key? if no I think we can remove this guard

jebentier (changes requested):

lib/declare_schema/model/habtm_model_shim.rb
yinonrousso: Oh gotchya, I'll go ahead and undo this

lib/declare_schema/model/index_definition.rb
yinonrousso: Nice! That separation definitely _makes this better_

lib/declare_schema/model/index_definition.rb
yinonrousso: ack 👍

lib/declare_schema/model/habtm_model_shim.rb
```diff
@@ -53,8 +57,7 @@ def _declared_primary_key
 
       def index_definitions_with_primary_key
         [
-          IndexDefinition.new(self, foreign_keys, unique: true, name: ::DeclareSchema::Model::IndexDefinition::PRIMARY_KEY_NAME),
-          IndexDefinition.new(self, foreign_keys.last) # not unique by itself; combines with primary key to be unique
```
yinonrousso, line 57: I think this was intentionally added in order to keep the `change_index` method in parity with what the actual active_record indexes look like. I added a workaround in fb7f371f9876dcdbb49e8925ebd9b6bd3ad07a51 but maybe we want to add this back and be explicit about this definition.

lib/declare_schema/model/index_definition.rb
```diff
@@ -33,12 +33,15 @@ def initialize(model, fields, **options)
...
         def for_model(model, old_table_name = nil)
           t = old_table_name || model.table_name
 
-          primary_key_columns = Array(model.connection.primary_key(t)).presence or
-            raise "could not find primary key for table #{t} in #{model.connection.columns(t).inspect}"
+          habtm_model = model.is_a?(DeclareSchema::Model::HabtmModelShim)
+
+          primary_key_columns = Array(model.connection.primary_key(t)).presence
+            primary_key_columns || habtm_model or
+              raise "could not find primary key for table #{t} in #{model.connection.columns(t).inspect}"
```
jebentier, line 44: Looks like there's an extra indentation here that we should remove
```suggestion
          primary_key_columns = Array(model.connection.primary_key(t)).presence
          primary_key_columns || habtm_model or
            raise "could not find primary key for table #{t} in #{model.connection.columns(t).inspect}"
```

lib/declare_schema/model/habtm_model_shim.rb
jebentier: Looking at the other fix you put in, this is much more explicit and can be more easily understood what is happening here. For that reason I think we should leave the old implementation over the more obscure `.pop` that was added in the linked commit.

jebentier (changes requested):
Changes are looking really good, just a couple updates and we should be good to go.

lib/declare_schema/model/index_definition.rb
yinonrousso: 💢👾

yinonrousso: fixed in 06cd051076355905fb5f9910c463c2c863fe303c

jebentier (approved):

PR #150, opened 2022-09-13
TECH-8177: add ability to use rake for migration generation for non-rails applications

jebentier:
I went with a quick and dirty solution here to the problem of generating migrations in a non-Rails application. I have a service, `fish` that is a Sinatra application using ActiveRecord and would like to be able to use declare schema to manage the database schema.

# Changelog

### Added
- Added a rake task definition that can be optionally included into a non-Rails project to generate
  schema migrations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ Inspired by [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 Note: this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.2.0] - Unreleased
+### Added
+- Added a rake task definition that can be optionally included into a non-Rails project to generate
+  schema migrations.
+
 ## [1.1.0] - 2022-07-22
 ### Changed
 - Fixed a bug where `DeclareSchema::Model::HabtmModelShim` `indexes` and `integer limits` were not being generated properly. Use `limit 8` for ids and primary composite key for habtm model.
@@ -229,6 +234,7 @@ using the appropriate Rails configuration attributes.
 ### Added
 - Initial version from https://github.com/Invoca/hobo_fields v4.1.0.
 
+[1.2.0]: https://github.com/Invoca/declare_schema/compare/v1.1.0...v1.2.0
 [1.1.0]: https://github.com/Invoca/declare_schema/compare/v1.0.2...v1.1.0
 [1.0.2]: https://github.com/Invoca/declare_schema/compare/v1.0.1...v1.0.2
 [1.0.1]: https://github.com/Invoca/declare_schema/compare/v1.0.0...v1.0.1
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ Note: this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0
 - Added a rake task definition that can be optionally included into a non-Rails project to generate
   schema migrations.
 
+### Fixed
+- Fixed a bug where not aliasing column names was allowing mysql to return them using their uppercase
+  variants
+
 ## [1.1.0] - 2022-07-22
 ### Changed
 - Fixed a bug where `DeclareSchema::Model::HabtmModelShim` `indexes` and `integer limits` were not being generated properly. Use `limit 8` for ids and primary composite key for habtm model.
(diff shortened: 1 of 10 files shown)


lib/declare_schema/rake.rb
```diff
@@ -0,0 +1,52 @@
...
+    puts "----------------------------------"
+
+    puts "\n---------- Down Migration --------"
+    puts down
+    puts "----------------------------------"
+
+    final_migration_name = default_migration_name
+    migration_template   = ERB.new(<<~EOF.chomp)
+      # frozen_string_literal: true
+      class <%= @migration_class_name %> < (ActiveRecord::Migration[4.2])
```
ColinDKelley, line 30: BTW, I wonder if the `[4.2]` will cause us problems someday. I guess they plan to support it forever...but still, we might do better with a more modern format at some point.

ColinDKelley: The parens here seem unnecessary:
```
< (ActiveRecord::Migration[4.2])
```
I'm guessing they were left over from when we had some ternary logic in there.

CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ Inspired by [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 Note: this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.2.0] - Unreleased
+### Added
+- Added a rake task definition that can be optionally included into a non-Rails project to generate
+  schema migrations.
```
ColinDKelley, line 10: Can you put a comment at the top of `lib/declare_schema/rake.rb` saying this is its (only) purpose? And add a bit to the README explaining how to use it with/without Rails?

lib/declare_schema/rake.rb
```diff
@@ -0,0 +1,52 @@
...
+
+    @up = "    #{up.strip.split("\n").join("\n    ")}"
+    @down = "    #{down.strip.split("\n").join("\n    ")}"
+    @migration_class_name = final_migration_name.camelize
+
+    File.write("db/migrate/#{Time.now.to_i}_#{final_migration_name.underscore}.rb", migration_template.result(binding))
+  end
+end
+
+def default_migration_name
```
ColinDKelley, line 48: It's surprising to me that we have logic here to generate a unique class name based on searching the old migrations, but just above we have code to choose a unique filename by adding the Time.now in front. Do we really need that if the name is already unique? Or is that just to get the files to sort in chronological order?

lib/declare_schema/rake.rb
jebentier: Good point. I went ahead and switched this to use the same template that's used by the generators (duplicated this originally to implement the rake task outside of the gem). Also removed the unnecessary `()`s. Agree we should be mindful of the rails version in the migrations so that we can upgrade and start using more current functionality for rails migrations as well.

a86eb21682e2e7cb3bcedbb860c38288f52bf11b

CHANGELOG.md
jebentier: Done 0b6ace6a3ca067c4110987ad58259ed44d933e06

lib/declare_schema/rake.rb
jebentier: Yea, this was also duplicated from when I was working on this from outside of the `declare_schema` project. I went ahead and switched the task to use the built in functionality to the gem with a slight change to avoid using the `Rails.root` method.

a86eb21682e2e7cb3bcedbb860c38288f52bf11b

lib/declare_schema/model/column.rb
```diff
@@ -132,7 +132,7 @@ def collation_and_charset_for_column(current_table_name, column_name)
         database_name = connection.current_database
 
         defaults = connection.select_one(<<~EOS)
-          SELECT C.character_set_name, C.collation_name
+          SELECT C.character_set_name as character_set_name, C.collation_name as collation_name
```
jebentier, line 135: @ColinDKelley Tacking on an additional bug fix that @cmparkinson found when playing with MySQL 8. Without the `as` aliasing of the column names, his MySQL configuration was returning these fields in their ALL_CAPS form. Now we're safe in both places we're querying `information_schema`.

ColinDKelley: Interesting! That's really odd. Glad you were able to get the fix in here.

ColinDKelley (approved):
Looks great!

PR #151, opened 2022-09-24
raise if Rails is defined yet Rails.application is nil

ColinDKelley:
### Fixed
- If Rails is defined, raise if `Rails.application` is `nil`.

### DIFF
--- a/lib/generators/declare_schema/migration/migrator.rb
+++ b/lib/generators/declare_schema/migration/migrator.rb
@@ -55,6 +55,7 @@ module Generators
         def load_rails_models
           ActiveRecord::Migration.verbose = false
           if defined?(Rails)
+            Rails.application or raise "Rails is defined, so Rails.application must be set"
             Rails.application.eager_load!
             Rails::Engine.subclasses.each(&:eager_load!)
           end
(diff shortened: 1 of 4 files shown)


jebentier (approved):

lib/generators/declare_schema/migration/migrator.rb
```diff
@@ -55,7 +55,7 @@ def initialize(renames: nil, &block)
         def load_rails_models
           ActiveRecord::Migration.verbose = false
           if defined?(Rails)
-            Rails.application.eager_load!
+            Rails.application&.eager_load!
```
jebentier, line 58: I don’t think we want to take on this edge case in the gem. We will introduce latent bug potential here by allowing a non-initialized rails environment to work here because this could easily be a boot order issue. Declare Schema should either work in an initialized rails environment or not have rails loaded. I would vote for raising an exception here that is explicit stating that rails is detected but not initialized and have the developer resolve the issue in their own application.

ColinDKelley: Ok, that sounds like a good compromise for now.

jebentier (approved):

PR #152, opened 2023-01-24
TECH-9308 Update README for belongs_to index behavior, ignore_tables …

robertp:


README.md
```diff
@@ -224,6 +224,39 @@ If your repo has a different command to run for migrations, you can configure it
 `DeclareSchema.db_migrate_command = "bundle exec rails db:migrate_immediate"`
 ```
 
+## The `belongs_to` Relation and Indexes
+By default, `declare_schema` creates an index for `belongs_to` relations. If an index
```
ColinDKelley, line 228: I'd start with a higher level sentence than this. Something like:
```
`declare_schema` intercepts the `belongs_to` macro to infer the foreign key column in a model that "points to" the primary key of the associated model. Note that `belongs_to` is outside the `declare_schema do` block.
```
I just remembered that we should have a separate section like this one for `has_and_belongs_to_many`. In that case `declare_schema` infers the intersection table with its foreign keys back to the two associated models.

ColinDKelley, line 228: Rather than "relations" I'd go with "associations". (I think "relation" is relational DB theory is actually the table, not the foreign keys across table.)

README.md
```diff
@@ -224,6 +224,39 @@ If your repo has a different command to run for migrations, you can configure it
...
+subsequently needs to be removed, it must be specifically updated in the relation statement.
```
ColinDKelley, line 229: I wouldn't focus on removing or adding. I'd just say "If this index is not desired, set `index: false`." We could cite the most common example for that, too, which is when a different index already covers this column at the front, as in:
```ruby
index [:other_table, :name]
```
Also, BTW, let's not use `belongs_to ... null: false` in the example, since Rails now has their own notation for this: `optional: false`. (`declare_schema` supports both, but just for backward-compatibility.)

README.md
```diff
@@ -224,6 +224,39 @@ If your repo has a different command to run for migrations, you can configure it
...
+```
+
+The schema should then be updated, e.g.:
+
+```shell
+bundle exec rails g declare_schema:migration
+```
+
+## Ignored Tables
+`declare_schema` by default ignores these tables:
```
ColinDKelley, line 243: I think an intro sentence would be good here as well. Something like
```
`declare_schema` can ignore any tables whose schema is managed elsewhere and not by the `declare_schema` gem.

ColinDKelley (changes requested):
BTW I see some tests failed. Probably just flukes since this is just changing a README!

README.md
robertp: Section rewritten, thank you. Please let me know if you'd like more in the new habtm section.

robertp: Right you are, that's loose terminology on my part.

README.md
robertp: Section rewritten

README.md
robertp: Section rewritten

ColinDKelley (approved):

PR #153, opened 2023-03-27
TECH-5882: renaming a foreign key currently generates incorrect migrations

gwynethallwright:
When a foreign key is renamed, `declare_schema` creates a migration that includes more than just a column rename. It also includes additions and removals of an index and foreign key. This PR gets rid of everything except the column rename.

### DIFF
--- a/lib/declare_schema/model/foreign_key_definition.rb
+++ b/lib/declare_schema/model/foreign_key_definition.rb
@@ -12,7 +12,7 @@ module DeclareSchema
 
       def initialize(model, foreign_key, **options)
         @model = model
-        @foreign_key = foreign_key.to_s.presence or raise ArgumentError "Foreign key cannot be empty: #{foreign_key.inspect}"
+        @foreign_key = foreign_key.to_s.presence or raise ArgumentError "Foreign key must not be empty: #{foreign_key.inspect}"
         @options = options
 
         @child_table_name = model.table_name # unless a table rename, which would happen when a class is renamed??
--- a/lib/declare_schema/model/foreign_key_definition.rb
+++ b/lib/declare_schema/model/foreign_key_definition.rb
@@ -12,7 +12,7 @@ module DeclareSchema
 
       def initialize(model, foreign_key, **options)
         @model = model
-        @foreign_key = foreign_key.to_s.presence
+        @foreign_key = foreign_key.to_s.presence or raise ArgumentError "Foreign key cannot be empty: #{foreign_key.inspect}"
         @options = options
 
         @child_table_name = model.table_name # unless a table rename, which would happen when a class is renamed??
--- a/lib/declare_schema/model/habtm_model_shim.rb
+++ b/lib/declare_schema/model/habtm_model_shim.rb
@@ -53,7 +53,7 @@ module DeclareSchema
 
       def index_definitions_with_primary_key
         [
-          IndexDefinition.new(self, foreign_keys, unique: true, name: "PRIMARY KEY"), # creates a primary composite key on both foreign keys
+          IndexDefinition.new(self, foreign_keys, unique: true, name: Model::IndexDefinition::PRIMARY_KEY_NAME), # creates a primary composite key on both foreign keys
           IndexDefinition.new(self, foreign_keys.last) # not unique by itself; combines with primary key to be unique
         ]
       end
--- a/lib/declare_schema/model/habtm_model_shim.rb
+++ b/lib/declare_schema/model/habtm_model_shim.rb
@@ -37,10 +37,6 @@ module DeclareSchema
         join_table
       end
 
-      def index_name
-        "index_#{table_name}_on_#{foreign_keys.first}_#{foreign_keys.last}"
-      end
-
       def field_specs
         foreign_keys.each_with_index.each_with_object({}) do |(v, position), result|
           result[v] = ::DeclareSchema::Model::FieldSpec.new(self, v, :bigint, position: position, null: false)
@@ -57,7 +53,7 @@ module DeclareSchema
 
       def index_definitions_with_primary_key
         [
-          IndexDefinition.new(self, foreign_keys, unique: true, name: index_name), # creates a primary composite key on both foreign keys
+          IndexDefinition.new(self, foreign_keys, unique: true, name: "PRIMARY KEY"), # creates a primary composite key on both foreign keys
           IndexDefinition.new(self, foreign_keys.last) # not unique by itself; combines with primary key to be unique
         ]
       end
--- a/lib/declare_schema/model/habtm_model_shim.rb
+++ b/lib/declare_schema/model/habtm_model_shim.rb
@@ -57,7 +57,7 @@ module DeclareSchema
 
       def index_definitions_with_primary_key
         [
-          IndexDefinition.new(self, foreign_keys, unique: true, name: index_name), # creates a primary composite key on both foriegn keys
+          IndexDefinition.new(self, foreign_keys, unique: true, name: index_name), # creates a primary composite key on both foreign keys
           IndexDefinition.new(self, foreign_keys.last) # not unique by itself; combines with primary key to be unique
         ]
       end
--- a/lib/declare_schema/model/index_definition.rb
+++ b/lib/declare_schema/model/index_definition.rb
@@ -33,14 +33,11 @@ module DeclareSchema
 
       class << self
         # extract IndexSpecs from an existing table
-        # always includes the PRIMARY KEY index for Model unless it is a HABTM Model.
         def for_model(model, old_table_name = nil)
           t = old_table_name || model.table_name
 
-          habtm_model = model.is_a?(DeclareSchema::Model::HabtmModelShim)
-
           primary_key_columns = Array(model.connection.primary_key(t)).presence
-          primary_key_columns || habtm_model or
+          primary_key_columns or
             raise "could not find primary key for table #{t} in #{model.connection.columns(t).inspect}"
 
           primary_key_found = false
@@ -50,14 +47,14 @@ module DeclareSchema
               i.columns == primary_key_columns && i.unique or
                 raise "primary key on #{t} was not unique on #{primary_key_columns} (was unique=#{i.unique} on #{i.columns})"
               primary_key_found = true
-            elsif i.columns == primary_key_columns && i.unique && !habtm_model
+            elsif i.columns == primary_key_columns && i.unique
               # skip this primary key index since we'll create it below, with PRIMARY_KEY_NAME
               next
             end
             new(model, i.columns, name: i.name, unique: i.unique, where: i.where, table_name: old_table_name)
           end.compact
 
-          if !primary_key_found && !habtm_model
+          if !primary_key_found
             index_definitions << new(model, primary_key_columns, name: PRIMARY_KEY_NAME, unique: true, where: nil, table_name: old_table_name)
           end
           index_definitions
--- a/lib/declare_schema/model/index_definition.rb
+++ b/lib/declare_schema/model/index_definition.rb
@@ -33,6 +33,7 @@ module DeclareSchema
 
       class << self
         # extract IndexSpecs from an existing table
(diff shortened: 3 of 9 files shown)


lib/generators/declare_schema/migration/migrator.rb
```diff
@@ -451,31 +451,68 @@ def change_indexes(model, old_table_name, to_rename)
...
+              to_rename[column]
+            end
+
+            indexes_to_add.each { |index_to_add|
+              if renamed_columns == index_to_add.columns
+                renamed_indexes_to_drop.append(index_to_drop)
+                renamed_indexes_to_add.append(index_to_add)
+              end
+            }
+          }
```
gwynethallwright, line 471: This is where I check if any add or drop index changes are due to a column rename. In this snippet, `columns` is an array of column names and `to_rename` is a hash that maps renaming changes.

lib/generators/declare_schema/migration/migrator.rb
```diff
@@ -451,31 +451,68 @@ def change_indexes(model, old_table_name, to_rename)
...
+
+          drop_indexes = (indexes_to_drop - renamed_indexes_to_drop).map do |i|
             ::DeclareSchema::SchemaChange::IndexRemove.new(new_table_name, i.columns, unique: i.unique, where: i.where, name: i.name)
           end
 
-          add_indexes = (model_indexes_without_primary_key - existing_indexes_without_primary_key).map do |i|
+          add_indexes = (indexes_to_add - renamed_indexes_to_add).map do |i|
```
gwynethallwright, line 477: Here is where I subtract out the indexes that I don't actually want to add. I do the same thing for the index removes.

lib/generators/declare_schema/migration/migrator.rb
```diff
@@ -451,31 +451,68 @@ def change_indexes(model, old_table_name, to_rename)
...
+
+          fks_to_drop.each { |fk_to_drop|
+            fks_to_add.each { |fk_to_add|
+              if fk_to_add.child_table_name == fk_to_drop.child_table_name &&
+                fk_to_add.parent_table_name == fk_to_drop.parent_table_name &&
+                fk_to_add.foreign_key == to_rename[fk_to_drop.foreign_key]
+
+                renamed_fks_to_drop.append(fk_to_drop)
+                renamed_fks_to_add.append(fk_to_add)
+              end
```
gwynethallwright, line 506: The approach for the foreign keys is similar. Something I'm looking for feedback on is this condition here: is this an acceptable necessary and sufficient condition for foreign key changes being due to a column rename?

lib/generators/declare_schema/migration/migrator.rb
```diff
@@ -451,31 +451,68 @@ def change_indexes(model, old_table_name, to_rename)
...
-          drop_indexes = (existing_indexes_without_primary_key - model_indexes_without_primary_key).map do |i|
+          indexes_to_drop = existing_indexes_without_primary_key - model_indexes_without_primary_key
+          indexes_to_add = model_indexes_without_primary_key - existing_indexes_without_primary_key
+
+          renamed_indexes_to_drop = []
+          renamed_indexes_to_add = []
+
+          indexes_to_drop.each { |index_to_drop|
+            renamed_columns = index_to_drop.columns.map do |column|
+              to_rename[column]
```
ColinDKelley, line 462: I'm assuming this hash lookup can sometimes return `nil`. Is that right? Will that cause a problem to have `nil` values in the array? If so, you could use `map_compact`.

lib/generators/declare_schema/migration/migrator.rb
```diff
@@ -451,31 +451,68 @@ def change_indexes(model, old_table_name, to_rename)
...
+            end
+
+            indexes_to_add.each { |index_to_add|
+              if renamed_columns == index_to_add.columns
```
ColinDKelley, line 466: These are both arrays, so comparing for equality will also check that the column order matches. Is there a chance that the set of columns matches, but the order doesn't (so `==` fails)? If there's a chance of that, you could always put them each into a `Set` first and then compare the sets.

lib/generators/declare_schema/migration/migrator.rb
```diff
@@ -451,31 +451,68 @@ def change_indexes(model, old_table_name, to_rename)
...
+          fks_to_add = model_fks - existing_fks
+
+          renamed_fks_to_drop = []
+          renamed_fks_to_add = []
+
+          fks_to_drop.each { |fk_to_drop|
+            fks_to_add.each { |fk_to_add|
+              if fk_to_add.child_table_name == fk_to_drop.child_table_name &&
+                fk_to_add.parent_table_name == fk_to_drop.parent_table_name &&
+                fk_to_add.foreign_key == to_rename[fk_to_drop.foreign_key]
```
ColinDKelley, line 502: I have a similar question about `nil` from the hash lookup. Can that happen? Any chance that the `foreign_key` method could return `nil` as well? If both of those happened, I'm guessing we wouldn't want to take the `==` path here.

lib/generators/declare_schema/migration/migrator.rb
ColinDKelley: I'm not certain! Might be worth talking through on Zoom. When in doubt, I hope test cases should enumerate the possibilities...but I know sometimes it's hard to know ahead of time the universe of cases to have tests around...

lib/generators/declare_schema/migration/migrator.rb
gwynethallwright: Good point. I'll look into it.

lib/generators/declare_schema/migration/migrator.rb
ColinDKelley: Sounds good. You could even use `.fetch` which takes a default value to use when the key is not found:
```
to_rename.fetch(column, column)
```

lib/generators/declare_schema/migration/migrator.rb
gwynethallwright: I don't have a conclusive answer to this yet. But from looking through the code a bit, I would hazard a guess that `fk_to_add.foreign_key` being `nil` would indicate a bug. Here is my reasoning:

1. We have `fks_to_add = model_fks - existing_fks`.
2. `model_fks` is generated from [here](https://github.com/Invoca/declare_schema/blob/b8bbf31f0a1df01ac61a3f4ac33de09ff252c394/lib/declare_schema/model/habtm_model_shim.rb#L73-L74). Trying to figure out if `foreign_keys` on that line could ever be `[]` was leading me down a rabbit hole, so this is where I'm missing clarity.
3. However, `existing_fks` seems a bit clearer. We generate it from the regex [here](https://github.com/Invoca/declare_schema/blob/b8bbf31f0a1df01ac61a3f4ac33de09ff252c394/lib/declare_schema/model/foreign_key_definition.rb#L33), which looks like it expects the `foreign_key` to be non-nil.
4. Since we subtract one array from the other, and since we expect the `foreign_key` to be non-nil for the part that we subtract, I'd guess that we expect it to be non-nil for the thing that we subtract it from.

We do have the `.presence` [here](https://github.com/Invoca/declare_schema/blob/b8bbf31f0a1df01ac61a3f4ac33de09ff252c394/lib/declare_schema/model/foreign_key_definition.rb#L15) though, so I'm a bit unsure. It might be worth throwing in something to prevent the `nil == nil` situation here just in case it did ever arise, even as a result of another bug.

lib/generators/declare_schema/migration/migrator.rb
gwynethallwright: I ended up going with the `Set` comparison. I'm not 100% certain that we need it, but the code is quite complex and I feel I could easily be overlooking something. I'm not aware of any reason using `Set` here would be a bad idea — and to me it seems like it would make the code a bit more resilient.

lib/generators/declare_schema/migration/migrator.rb
gwynethallwright: I ended up using `.fetch`. Thanks!

gwynethallwright:
New changes are working as expected and I tested a handful of non-rename migrations to verify that nothing broke. But still open to discussing this more or using another approach. I definitely don't have much familiarity with the code I'm changing.

jebentier (changes requested):

lib/declare_schema/model/foreign_key_definition.rb
```diff
@@ -12,7 +12,7 @@ class ForeignKeyDefinition
 
       def initialize(model, foreign_key, **options)
         @model = model
-        @foreign_key = foreign_key.to_s.presence
+        @foreign_key = foreign_key.to_s.presence or raise ArgumentError "Foreign key cannot be empty: #{foreign_key.inspect}"
```
ColinDKelley, line 15: How about "must not be empty"? I recall that being preferred grammatically.

lib/declare_schema/model/habtm_model_shim.rb
```diff
@@ -37,10 +37,6 @@ def table_name
         join_table
       end
 
-      def index_name
-        "index_#{table_name}_on_#{foreign_keys.first}_#{foreign_keys.last}"
-      end
-
```
ColinDKelley, line 43: Hurray! 💯

lib/declare_schema/model/index_definition.rb
```diff
@@ -33,14 +33,11 @@ def initialize(model, fields, **options)
 
       class << self
         # extract IndexSpecs from an existing table
-        # always includes the PRIMARY KEY index for Model unless it is a HABTM Model.
```
ColinDKelley, line 36: I think this part of the note is worth keeping:
```
# includes the PRIMARY KEY index
```

lib/declare_schema/model/index_definition.rb
```diff
@@ -33,14 +33,11 @@ def initialize(model, fields, **options)
...
         def for_model(model, old_table_name = nil)
           t = old_table_name || model.table_name
 
-          habtm_model = model.is_a?(DeclareSchema::Model::HabtmModelShim)
-
           primary_key_columns = Array(model.connection.primary_key(t)).presence
-          primary_key_columns || habtm_model or
+          primary_key_columns or
             raise "could not find primary key for table #{t} in #{model.connection.columns(t).inspect}"
```
ColinDKelley, line 41: These might fit on one line now.

lib/generators/declare_schema/migration/migrator.rb
```diff
@@ -521,6 +507,22 @@ def change_foreign_key_constraints(model, old_table_name, to_rename)
           [drop_fks + add_fks]
         end
 
+        def foreign_key_changes_due_to_column_renames(fks_to_drop, fks_to_add, to_rename)
+          fks_to_drop.each_with_object([[], []]) do |fk_to_drop, (renamed_fks_to_drop, renamed_fks_to_add)|
+            fk_to_add = fks_to_add.find do |fk_to_add|
+              fk_to_add.foreign_key.nil? and raise "Foreign key is not allowed to be nil for #{fk_to_add.inspect}"
```
ColinDKelley, line 513: I think you can drop this assertion now that foreign keys are prohibited from being `nil?`.

lib/declare_schema/model/index_definition.rb
gwynethallwright: They do! I'll fix it.

lib/generators/declare_schema/migration/migrator.rb
gwynethallwright: Yes, that makes sense.

lib/declare_schema/model/foreign_key_definition.rb
gwynethallwright: Yes, using "must not" makes it sound like more of an imperative. I'll change the wording.

gwynethallwright:
Here is a link to a web build using the latest version of the gem: https://buildkite.com/invoca/web/builds/53821.

ColinDKelley (approved):
Looks good! Thanks for being so thorough.

Issue #154, opened 2023-05-31
Index names are not unique


garvank:
I have a rails application using sqlite locally and postgres in production. 

#### Issue
When trying to create an index on a field, for example: "user_id on comments" and "user_id on assignments" my generated migrations are unable to run due to an error stating that "index :on_user_id already exists". 

#### STR

1. Have 3 models, for example `Comment`, `Assignment`, and `User`
2. `Comment` and `Assignement` both `belong_to :user`
3. When running the second generated migration, see the error

#### Workaround
Rename `:on_user_id` to `:user_id_on_comments` in the migration file

#### Expected Behavior
The generated migrations generate unique index names and can be run successfully

ColinDKelley:
Thanks for letting us know about this on Postgres. It was my mistake to assume MySQL uniqueness rules (where index names are unique per table) vs having index names be globally unique. The incentive there was that MySQL has a hard limit of 64 characters for index names and we were blowing up there sometimes.

Here's an idea for a compromise:
1. Allow the max index name to be configured. Default this to 64, but you could set it to `nil` meaning unlimited, for example if you'll only be deploying to Postgres.
2. Change the [default index name generator](https://github.com/Invoca/declare_schema/blob/7a0a75966bc23c1ebd9ee3140e6303817b068f49/lib/declare_schema/model/index_definition.rb#L60-L62) to honor the configured length by trying these conventions in turn until the first is found that fits within the configured limit, if any:
- `"index_#{table_name}_on_#{column_names.join('_and_')}"`  <--- the Rails convention
- `"#{table_name}__#{column_names.join('__')}"`
- raise an exception indicating that an explicit `name:` must be chosen to stay within the configured limit

Sound good?

ColinDKelley:
^ @garvank Sound good?

garvank:
@ColinDKelley Configuration and docs describing them sound good to me! The other thing that comes to mind is that maybe it _must_ be configured or else raise an error.

ColinDKelley:
@garvank I decided not to require configuration. I just default the config to 64 characters since that's what you need if you want the code to work with all possible database types. For Postgres you can set the config to `nil` if the version you're using is truly unlimited:
```ruby
DeclareSchema.max_index_and_constraint_name_length = nil
```
I've published a pre-release version with this change. Can you try it out and report back on how it works for you? The version is `declare_schema 1.3.0.colin.1`.

[...]

PR #155, opened 2023-07-03
Issue #154: index names not globally unique

ColinDKelley:
This addresses issue #154 by putting the index name code back as it was, generating globally unique names, since those are required for some database like Postgres.

In addition:

### Added
- Added `DeclareSchema.max_index_and_constraint_name_length` with default of 64.

In the event that the full name like `index_users_on_last_name_and_first_name_and_middle_name` is too long, a shorter name will be tried, like `users__last_name_first_name_middle_name`.


spec/lib/declare_schema_spec.rb
```diff
@@ -5,97 +5,127 @@
     subject { described_class.default_charset }
 
     context 'when not explicitly set' do
-      it { should eq("utf8mb4") }
+      it { is_expected.to eq("utf8mb4") }
```
ColinDKelley, line 8: We're standardizing on `is_expected.to`.

ishakun (approved):

lib/declare_schema/model/index_definition.rb
```diff
@@ -57,8 +56,25 @@ def for_model(model, old_table_name = nil)
...
+        end
+
+        private
+
+        def long_index_name(table_name, columns)
+          "index_#{table_name}_on_#{Array(columns).join("_and_")}"
+        end
+
+        def short_index_name(table_name, columns)
+          "#{table_name}__#{Array(columns).join("_")}"
```
garvank, line 77: Pretty sure prefixing with the `table_name` is what fixed it for me`. Thanks!

garvank (approved):

PR #156, opened 2023-07-06
TECH-11529: Allow IndexDefinitions for HABTM models defined without a…

dcaddell:
… primary key

Note for reviewer: Will bump the version and update the CHANGELOG if this fix looks reasonable.

### Fixed
- Fixed a bug where a HABTM model without a primary key would cause the migration generator to fail.

### DIFF
--- a/lib/declare_schema/model/habtm_model_shim.rb
+++ b/lib/declare_schema/model/habtm_model_shim.rb
@@ -37,6 +37,10 @@ module DeclareSchema
         join_table
       end
 
+      def index_name
+        "index_#{table_name}_on_#{foreign_keys.first}_#{foreign_keys.last}"
+      end
+
       def field_specs
         foreign_keys.each_with_index.each_with_object({}) do |(v, position), result|
           result[v] = ::DeclareSchema::Model::FieldSpec.new(self, v, :bigint, position: position, null: false)
@@ -53,7 +57,7 @@ module DeclareSchema
 
       def index_definitions_with_primary_key
         [
-          IndexDefinition.new(self, foreign_keys, unique: true, name: Model::IndexDefinition::PRIMARY_KEY_NAME), # creates a primary composite key on both foreign keys
+          IndexDefinition.new(self, foreign_keys, unique: true, name: index_name), # creates a primary composite key on both foreign keys
           IndexDefinition.new(self, foreign_keys.last) # not unique by itself; combines with primary key to be unique
         ]
       end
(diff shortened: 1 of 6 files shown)


ColinDKelley:
Interesting that we had HABTM models without a primary key. Can you point me to one of those? I _guess_ it's ok since there's no actual Model over them... I'm unsure on the Rails best practice.

ColinDKelley:
Oh, this seems relevant: https://rails.lighthouseapp.com/projects/8994-ruby-on-rails/tickets/2086-patch-primary-key-on-habtm-join-table-now-raises-an-exception

> Join tables for has_and_belongs_to_many relationships are not supposed to have primary keys. This is mentioned in a few places in the documentation, but remains an easy "gotcha". Unless you know this rule, you can create several habtm relationships before the process eventually fails in a cryptic manner. It's very likely that your tests will not catch this, since it does not fail right away.

Very interesting, if true!

ColinDKelley:
The ideal design from a SQL perspective would be for the HABTM tables to have a composite unique index on the pair of foreign keys. I read some discussion of that. There's been a composite-primary-key gem for more than a decade that seemed to be gaining traction to get pulled into Rails...

dcaddell:
Awesome find @ColinDKelley thanks for stopping us from falling into that trap.

> The ideal design from a SQL perspective would be for the HABTM tables to have a composite unique index on the pair of foreign keys. I read some discussion of that. There's been a composite-primary-key gem for more than a decade that seemed to be gaining traction to get pulled into Rails...

Yeah, creating the HABTM tables with a composite unique index on the foreign keys seems like the only other reasonable option and declare_schema is already behaving that way. I think the changes to habtm_model_shim and index_definition from https://github.com/Invoca/declare_schema/pull/153/files had made me think we wanted to go the other way. It'll be pretty simple to revert the HABTM-focused changes from that PR.

I don't think the composite-primary-key gem would help us here since the tables don't have a primary key. Are you thinking that using it might solve the usability issues caused by using a composite primary key in a HABTM table?

ColinDKelley:
> I don't think the composite-primary-key gem would help us here since the tables don't have a primary key. 

@dcaddell You pointed out that the Rails `create_table` API lets us create a composite primary key, though. Is that sufficient?

And I'm not certain if the code has settled down and you're ready for a review, so please LMK.

ColinDKelley:
@dcaddell I keep getting reminders about this PR. Do you think it will move forward soon? If not, no worries--I can untag myself as a reviewer and then you can retag me when ready.

dcaddell:
Sorry about that @ColinDKelley  - I'll want to work this sometime next week and probably after your refactor gets merged. I'll set it back to a draft and remove the reviewers

dcaddell:
Hmm it won't let me untag octothorpe so I'm going to close it for now. Not abandoning the work but don't want it to keep pinging folks.

PR #157, opened 2023-07-11
non-production: move sort into HabtmModelShim#initialize; omit connection and foreign_key_classes

ColinDKelley:
Refactor `HabtmModelShim`:
- move sorting into `initialize`
- omit `connection` and `foreign_key_classes` (replace with `table_names`)

Refactor `IndexSpecification`:
- use explicit keyword args rather than `**options`
- instead of `model`, just pass in `table_name` arg (this makes the `table_name:` keyword arg unnecessary)
- address TODO by changing `fields` to `columns` (but leave alias the other way)

Refactor `ForeignKeyDefinition`:
- use explicit keyword args rather than `**options`
- instead of `model`, just pass in `foreign_key_column`, `constraint_name:`.

### DIFF
--- a/lib/declare_schema/model.rb
+++ b/lib/declare_schema/model.rb
@@ -49,11 +49,14 @@ module DeclareSchema
     end
 
     module ClassMethods
-      def index(fields, **options)
+      def index(fields, name: nil, allow_equivalent: false, unique: false, where: nil)
         # make index idempotent
         index_fields_s = Array.wrap(fields).map(&:to_s)
         unless index_definitions.any? { |index_spec| index_spec.fields == index_fields_s }
-          index_definitions << ::DeclareSchema::Model::IndexDefinition.new(self, fields, **options)
+          index_definitions << ::DeclareSchema::Model::IndexDefinition.new(
+            table_name, fields,
+            name: name, allow_equivalent: allow_equivalent, unique: unique, where: where
+          )
         end
       end
 
@@ -208,7 +211,7 @@ module DeclareSchema
       end
 
       def _rails_default_primary_key
-        ::DeclareSchema::Model::IndexDefinition.new(self, [_declared_primary_key.to_sym], unique: true, name: DeclareSchema::Model::IndexDefinition::PRIMARY_KEY_NAME)
+        ::DeclareSchema::Model::IndexDefinition.new(table_name, [_declared_primary_key], unique: true, name: DeclareSchema::Model::IndexDefinition::PRIMARY_KEY_NAME)
       end
 
       # Declares the "foo_type" field that accompanies the "foo_id"
--- a/lib/declare_schema/model.rb
+++ b/lib/declare_schema/model.rb
@@ -49,19 +49,21 @@ module DeclareSchema
     end
 
     module ClassMethods
-      def index(fields, name: nil, allow_equivalent: false, unique: false, where: nil)
-        # make index idempotent
-        index_fields_s = Array.wrap(fields).map(&:to_s)
-        unless index_definitions.any? { |index_spec| index_spec.fields == index_fields_s }
-          index_definitions << ::DeclareSchema::Model::IndexDefinition.new(
-            table_name, fields,
+      def index(columns, name: nil, allow_equivalent: false, unique: false, where: nil)
+        index_definition =
+          ::DeclareSchema::Model::IndexDefinition.new(
+            table_name, columns,
             name: name, allow_equivalent: allow_equivalent, unique: unique, where: where
           )
+
+        # add idempotently
+        unless index_definitions.any? { |index_def| index_def.equivalent?(index_definition) }
+          index_definitions << index_definition
         end
       end
 
-      def primary_key_index(*fields)
-        index(fields.flatten, unique: true, name: ::DeclareSchema::Model::IndexDefinition::PRIMARY_KEY_NAME)
+      def primary_key_index(*columns)
+        index(columns.flatten, unique: true, name: ::DeclareSchema::Model::IndexDefinition::PRIMARY_KEY_NAME)
       end
 
       def constraint(fkey, **options)
--- a/lib/declare_schema/model.rb
+++ b/lib/declare_schema/model.rb
@@ -52,8 +52,8 @@ module DeclareSchema
       def index(columns, name: nil, allow_equivalent: false, unique: false, where: nil)
         index_definition =
           ::DeclareSchema::Model::IndexDefinition.new(
-            table_name, columns,
-            name: name, allow_equivalent: allow_equivalent, unique: unique, where: where
+            columns,
+            name: name, table_name: table_name, allow_equivalent: allow_equivalent, unique: unique, where: where
           )
 
         # add idempotently
@@ -213,7 +213,7 @@ module DeclareSchema
       end
 
       def _rails_default_primary_key
-        ::DeclareSchema::Model::IndexDefinition.new(table_name, [_declared_primary_key], unique: true, name: DeclareSchema::Model::IndexDefinition::PRIMARY_KEY_NAME)
+        ::DeclareSchema::Model::IndexDefinition.new([_declared_primary_key], name: DeclareSchema::Model::IndexDefinition::PRIMARY_KEY_NAME, unique: true)
       end
 
       # Declares the "foo_type" field that accompanies the "foo_id"
@@ -284,15 +284,14 @@ module DeclareSchema
         end
       end
 
-      def _add_index_for_field(name, args, options)
-        if (to_name = options.delete(:index))
-          index_opts =
-            {
-              unique: args.include?(:unique) || options.delete(:unique)
-            }
-          # support index: true declaration
-          index_opts[:name] = to_name unless to_name == true
-          index(name, **index_opts)
+      def _add_index_for_field(column_name, args, options)
+        if (index_name = options.delete(:index))
+          if index_name == true
+            index_name = nil # index: true means generate default name
+          end
+          unique = args.include?(:unique) || options.delete(:unique) || false
+
+          index([column_name], unique: unique, name: index_name)
         end
       end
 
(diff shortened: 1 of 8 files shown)


lib/declare_schema/model/habtm_model_shim.rb
```diff
@@ -5,28 +5,21 @@ module Model
...
-            [refl.foreign_key.to_s, refl.active_record],
-            [refl.association_foreign_key.to_s, refl.class_name.constantize]
-          ].sort { |a, b| a.first <=> b.first }
-          foreign_keys = foreign_keys_and_classes.map(&:first)
-          foreign_key_classes = foreign_keys_and_classes.map(&:last)
-          # this may fail in weird ways if HABTM is running across two DB connections (assuming that's even supported)
-          # figure that anybody who sets THAT up can deal with their own migrations...
-          connection = refl.active_record.connection
-
-          new(join_table, foreign_keys, foreign_key_classes, connection)
```
ColinDKelley, line 19: Since Rails requires the HABTM tables names be sorted, that should be in `initialize`.

lib/declare_schema/model/habtm_model_shim.rb
```diff
@@ -5,28 +5,21 @@ module Model
...
+          new(refl.join_table, [refl.foreign_key, refl.association_foreign_key],
+                               [refl.active_record.table_name, refl.class_name.constantize.table_name])
         end
       end
 
-      attr_reader :join_table, :foreign_keys, :foreign_key_classes, :connection
+      attr_reader :join_table, :foreign_keys, :table_names
```
ColinDKelley, line 13: No need to keep the `foreign_key_classes`--just the `table_names`. And no need for the `connection`; it was unused.

lib/declare_schema/model/index_definition.rb
```diff
@@ -103,7 +100,7 @@ def equivalent?(rhs)
       end
 
       def with_name(new_name)
-        self.class.new(@model, @fields, table_name: @table_name, index_name: @index_name, unique: @unique, name: new_name)
+        self.class.new(@table_name, @columns, unique: @unique, name: new_name)
```
dcaddell, line 103: I think we should also bring the `allow_equivalent` and `where` options along when using this method.

lib/declare_schema/model/habtm_model_shim.rb
```diff
@@ -5,28 +5,21 @@ module Model
...
+        foreign_keys.is_a?(Array) && foreign_keys.size == 2 or
+          raise ArgumentError, "foreign_keys must be <Array[2]>; got #{foreign_keys.inspect}"
+        table_names.is_a?(Array) && table_names.size == 2 or
+          raise ArgumentError, "table_names must be <Array[2]>; got #{table_names.inspect}"
         @join_table = join_table
-        @foreign_keys = foreign_keys
-        @foreign_key_classes = foreign_key_classes
-        @connection = connection
+        @foreign_keys = foreign_keys.sort
+        @table_names = @foreign_keys == foreign_keys ? table_names : table_names.reverse
```
dcaddell, line 22: I think it's a little confusing for there to be both a table_name and table_names defined since they aren't related to each other. Maybe this could be `parent_table_names` or `foreign_key_table_names` to be explicit about the difference?

spec/lib/declare_schema/model/index_definition_spec.rb
```diff
@@ -81,7 +82,7 @@ class IndexDefinitionCompoundIndexModel < ActiveRecord::Base
       end
 
       describe 'for_model' do
-        subject { described_class.for_model(model_class) }
+          subject { described_class.for_model(model_class, nil) }
```
dcaddell, line 85: ⛏️  extra spacing

dcaddell (changes requested):
This looks awesome, especially for some random midnight coding lol. I have a couple minor suggestions but this is a really helpful refactor so thanks for knocking it out.

lib/declare_schema/model/index_definition.rb
ColinDKelley: Agreed. I'd already added `where` support and just now I added `allow_equivalent`. Although we've agreed on the need to redo the "index changes on column renames" ticket that introduced this method...and I don't expect this method will be needed after that.

lib/declare_schema/model/habtm_model_shim.rb
```diff
@@ -5,28 +5,21 @@ module Model
...
-            [refl.association_foreign_key.to_s, refl.class_name.constantize]
-          ].sort { |a, b| a.first <=> b.first }
-          foreign_keys = foreign_keys_and_classes.map(&:first)
-          foreign_key_classes = foreign_keys_and_classes.map(&:last)
-          # this may fail in weird ways if HABTM is running across two DB connections (assuming that's even supported)
-          # figure that anybody who sets THAT up can deal with their own migrations...
-          connection = refl.active_record.connection
-
-          new(join_table, foreign_keys, foreign_key_classes, connection)
+          new(refl.join_table, [refl.foreign_key, refl.association_foreign_key],
```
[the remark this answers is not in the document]
ColinDKelley: I don't think those were ever needed. I recalled (and also confirmed) that the Rails Reflection class already stores those as strings.

lib/declare_schema/model/index_definition.rb
```diff
@@ -5,61 +5,58 @@ module Model
...
       PRIMARY_KEY_NAME = "PRIMARY"
 
-      def initialize(model, fields, **options)
-        @model = model
-        @table = options.delete(:table_name) || model.table_name
-        @fields = Array.wrap(fields).map(&:to_s)
-        @explicit_name = options[:name] unless options.delete(:allow_equivalent)
-        @name = options.delete(:name) || self.class.default_index_name(@table, @fields)
-        @unique = options.delete(:unique) || name == PRIMARY_KEY_NAME || false
+      def initialize(table_name, columns, name: nil, allow_equivalent: false, unique: false, where: nil)
```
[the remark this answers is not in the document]
ColinDKelley: Yeah, it's a huge benefit in Ruby to get away from unconstrained **options hashes in favor of concrete keyword arg lists. I've been pushing these concrete keyword arg lists up the stack as I go. For example here:
https://github.com/Invoca/declare_schema/pull/157/files/c7d38deb36deac229586378cce13d6d0154c37e1#diff-03410725437d96a040256756bb77266e9898f56a21f8e666e524b766f5ca74eaR52

lib/declare_schema/model/habtm_model_shim.rb
ColinDKelley: Good call. Updated here: https://github.com/Invoca/declare_schema/pull/157/commits/0c6cb41f88c5f30eb1fe84ad40e52a6418f913cb

lib/declare_schema/model.rb
```diff
@@ -49,22 +49,34 @@ def self.inherited(klass)
...
-      def index(fields, **options)
-        # make index idempotent
-        index_fields_s = Array.wrap(fields).map(&:to_s)
-        unless index_definitions.any? { |index_spec| index_spec.fields == index_fields_s }
-          index_definitions << ::DeclareSchema::Model::IndexDefinition.new(self, fields, **options)
+      def index(columns, name: nil, allow_equivalent: false, unique: false, where: nil)
+        index_definition =
+          ::DeclareSchema::Model::IndexDefinition.new(
+            columns,
+            name: name, table_name: table_name, allow_equivalent: allow_equivalent, unique: unique, where: where
```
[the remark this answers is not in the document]
ColinDKelley: um...

lib/declare_schema/model/habtm_model_shim.rb
```diff
@@ -5,28 +5,21 @@ module Model
...
-      def initialize(join_table, foreign_keys, foreign_key_classes, connection)
+      def initialize(join_table, foreign_keys, parent_table_names)
+        foreign_keys.is_a?(Array) && foreign_keys.size == 2 or
+          raise ArgumentError, "foreign_keys must be <Array[2]>; got #{foreign_keys.inspect}"
+        parent_table_names.is_a?(Array) && parent_table_names.size == 2 or
+          raise ArgumentError, "parent_table_names must be <Array[2]>; got #{parent_table_names.inspect}"
         @join_table = join_table
-        @foreign_keys = foreign_keys
-        @foreign_key_classes = foreign_key_classes
-        @connection = connection
```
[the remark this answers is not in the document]
ColinDKelley: > Models are still expected to implement connection in a few places.

True. I'm just trying to chase away dependencies in general, when I can. Fewer implicit dependencies is always an improvement...unless it causes the explicit dependencies to get annoying.

It is possible to use `ActiveRecord::Base.connection`, but in the general case that may not work, because Rails allows each model to be stored in a different database and hence need a different connection. I don't think that's very commonly used, but it definitely does get some use. I'm also not certain how well `declare_schema` would work straddling multiple databases like.

Still, to answer this question:

>  Definitely would be interested in learning about the practical different between ActiveRecord::Base.connection and refl.active_record.connection.

The one in ActiveRecord::Base is the "main" connection. Models typically share that same connection pool, but they are allowed to specify a different connection spec (and hence connection pool) if they want to, so that a subset of the models might live in a separate DB. This support was added to Rails 6 and roughly encompasses what we've been doing with ActiveTableSet. So we think we can probably phase out ATS.

So best practice is:
1. Use the connection from the model you're accessing.
2. If you have general code that isn't tied to any particular model, use ActiveRecord::Base.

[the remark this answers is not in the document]
dcaddell: Awesome thanks for that explanation that's really helpful.

So in this case, since the `HabtmModelShim` doesn't include `ActiveRecord::ConnectionHandling` we still need to copy/store that connection from the reflection right?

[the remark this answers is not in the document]
ColinDKelley: @dcaddell Yes. In the latest feature branch I decided to [make that explicit, with a keyword argument](https://github.com/Invoca/declare_schema/pull/162/commits/257883c84d2180acea5f2be502c6ccb05b1965fb). I like how clear and non-error-prone that is. It was also really easy to TDD.

PR #158, opened 2023-10-11
issue #154: shorten table name with sha256

ColinDKelley:
A bug came up when the table name is so long that the short default index name is still too long. To address this, we now fall back to the prefix of the table name, plus up to 4 characters of its sha256. [Slack thread](https://invoca.slack.com/archives/CRGB17J59/p1697040014363769?thread_ts=1697037804.977329&cid=CRGB17J59).

### DIFF
--- a/lib/declare_schema/model/index_definition.rb
+++ b/lib/declare_schema/model/index_definition.rb
@@ -66,7 +66,7 @@ module DeclareSchema
               break index_name
             end
           end or raise IndexNameTooLongError,
-                       "Index '#{index_name}' exceeds configured limit of #{DeclareSchema.max_index_and_constraint_name_length} characters."
+                       "Default index name '#{index_name}' exceeds configured limit of #{DeclareSchema.max_index_and_constraint_name_length} characters. Use the `name:` option to give it a shorter name, or adjust DeclareSchema.max_index_and_constraint_name_length if you know your database can accept longer names."
         end
 
         private
--- a/lib/declare_schema/model/index_definition.rb
+++ b/lib/declare_schema/model/index_definition.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'digest/sha2'
+
 module DeclareSchema
   module Model
     class IndexDefinition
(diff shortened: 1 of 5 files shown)


spec/lib/declare_schema/model/index_definition_spec.rb
```diff
@@ -133,6 +133,46 @@ class IndexDefinitionCompoundIndexModel < ActiveRecord::Base
...
+          44 => 'user_d4814__last_name_first_name_middle_name',
+          45 => 'user_do4814__last_name_first_name_middle_name',
+          46 => 'user_dom4814__last_name_first_name_middle_name',
+          47 => 'user_doma4814__last_name_first_name_middle_name',
+          48 => 'user_domai4814__last_name_first_name_middle_name',
+          49 => 'user_domain4814__last_name_first_name_middle_name',
+          50 => 'user_domains4814__last_name_first_name_middle_name',
+          51 => 'user_domains_4814__last_name_first_name_middle_name',
+          52 => 'user_domains_extra__last_name_first_name_middle_name',
+        }.each do |len, index_name|
```
dcaddell, line 159: Nit: It'd be nice to also show the `len == 53` case here to prove that it behaves the same as `len == 52`. Not strictly necessary with the other tests but I like how it illustrates the point.

lib/declare_schema/model/index_definition.rb
```diff
@@ -69,12 +71,30 @@ def default_index_name(table, fields)
...
         end
 
         def short_index_name(table_name, columns)
-          "#{table_name}__#{Array(columns).join("_")}"
+          columns_suffix = "__" + Array(columns).join('_')
+          if DeclareSchema.max_index_and_constraint_name_length.nil?
+            table_name + columns_suffix
+          else
+            max_name_len = [DeclareSchema.max_index_and_constraint_name_length - columns_suffix.length, 0].max
+            shorten_name(table_name, max_name_len) + columns_suffix
```
dcaddell, line 96: If this is still too long, are we ok with the expectation that users need to declare a name?

lib/declare_schema/model/index_definition.rb
```diff
@@ -69,12 +71,30 @@ def default_index_name(table, fields)
...
 
+        SHA_SUFFIX_LENGTH = 4
+
+        def shorten_name(name, max_len)
+          if name.size <= max_len
+            name
+          else
+            name_prefix = name.first(max_len >= SHA_SUFFIX_LENGTH*2 ? (max_len - SHA_SUFFIX_LENGTH) : ((max_len + 1)/2))
+            sha = Digest::SHA256.hexdigest(name)
+            (name_prefix + sha).first(max_len)
```
dcaddell, line 82: Tricky to understand what this will look like just by inspecting the code. Really appreciate the thorough test 💯

dcaddell: Also, what would you think about a warning message if we have to truncate the unique name? It might be nice to point people towards the `name` option of the index dsl.

dcaddell (commented):
This approach seems good to me, just have a couple minor questions on usability.

spec/lib/declare_schema/model/index_definition_spec.rb
ColinDKelley: Added: https://github.com/Invoca/declare_schema/pull/158/commits/772dd90cc86b668eb2f412da3e8a51a74e0ccf64

lib/declare_schema/model/index_definition.rb
ColinDKelley: Yes, I think that's the best we can do.

lib/declare_schema/model/index_definition.rb
ColinDKelley: Good point. I was just thinking that too! https://github.com/Invoca/declare_schema/pull/158/commits/9592b0ed42fc11733dcf4e0effe60b203ffab46d

dcaddell (approved):
lgtm!

lib/declare_schema/model/index_definition.rb
ColinDKelley: ...especially now that the error message suggests how to fix it. :)

abhig10 (approved):
LGTM. Thank you for the quick turnaround.

PR #162, opened 2024-01-16
TECH-13204/support partial (aka prefix) indexes

ColinDKelley:
### Added
- Added support for partial indexes with `length:` option.
### Changed
- Deprecate `index: 'name'` and `unique: true|false` in favor of `index: { name: 'name', unique: true|false }`.

Note: README refactor is entirely over here, since it was so interleaved.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ Inspired by [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 Note: this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [1.4] - Unreleased
+## [1.4.0] - 2024-01-24
 ### Added
 - Added support for partial indexes with `length:` option.
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ Inspired by [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 Note: this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.4] - Unreleased
+### Added
+- Added support for partial indexes with `length:` option.
+### Changed
+- Deprecate index: 'name' and unique: true|false in favor of index: { name: 'name', unique: true|false }.
+
 ## [1.3.6] - 2024-01-22
 ### Fixed
 - Add missing commits around connection: and Array check for composite declared primary key.
--- a/lib/declare_schema/model/index_definition.rb
+++ b/lib/declare_schema/model/index_definition.rb
@@ -18,7 +18,7 @@ module DeclareSchema
 
       def initialize(columns, table_name:, name: nil, allow_equivalent: false, unique: false, where: nil, length: nil)
         @table_name = table_name
-        @name = name || self.class.default_index_name(table_name, columns)
+        @name = (name || self.class.default_index_name(table_name, columns)).to_s
         @columns = Array.wrap(columns).map(&:to_s)
         @explicit_name = @name if !allow_equivalent
         unique.in?([false, true]) or raise ArgumentError, "unique must be true or false: got #{unique.inspect}"
--- a/lib/declare_schema/model/index_definition.rb
+++ b/lib/declare_schema/model/index_definition.rb
@@ -7,7 +7,7 @@ module DeclareSchema
     class IndexDefinition
       include Comparable
 
-      OPTIONS = [:name, :unique, :where].freeze
+      OPTIONS = [:name, :unique, :where, :length].freeze
       attr_reader :columns, :explicit_name, :table_name, *OPTIONS
 
       alias fields columns # TODO: change callers to use columns. -Colin
@@ -16,7 +16,7 @@ module DeclareSchema
 
       PRIMARY_KEY_NAME = "PRIMARY"
 
-      def initialize(columns, table_name:, name: nil, allow_equivalent: false, unique: false, where: nil)
+      def initialize(columns, table_name:, name: nil, allow_equivalent: false, unique: false, where: nil, length: nil)
         @table_name = table_name
         @name = name || self.class.default_index_name(table_name, columns)
         @columns = Array.wrap(columns).map(&:to_s)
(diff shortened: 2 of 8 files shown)


spec/lib/declare_schema/migration_generator_spec.rb
```diff
@@ -391,6 +391,7 @@ class Advert < ActiveRecord::Base
 
       Advert.field_specs.delete(:c_id)
       Advert.index_definitions.delete_if { |spec| spec.fields == ["c_id"] }
+      Advert.constraint_definitions.delete_if { |spec| spec.foreign_key_column == "c_id" }
```
ColinDKelley, line 394: Because this was missing, it was leaving a foreign key definition around, which added extra junk to the migrations below.

lib/declare_schema/model/index_definition.rb
```diff
@@ -51,11 +57,20 @@ def for_table(table_name, ignore_indexes, connection)
                 raise "primary key on #{table_name} was not unique on #{primary_key_columns} (was unique=#{index.unique} on #{index.columns})"
               primary_key_found = true
             end
-            new(index.columns, name: index.name, unique: index.unique, where: index.where)
+            length =
+              case lengths = index.lengths
+              when {}
+                nil
+              when Hash
+                lengths.size == 1 ? lengths.values.first : lengths
```
jebentier, line 65: ❓ What are the different hash shapes that `index.lengths` can take? This contract we're building feels brittle to me, so I'm curious if we can harden it with more strict expectations.

ColinDKelley: It appears to always return a Hash. (Not what I expected at first.) So a fresh cut at the code could be much cleaner:
```ruby
lengths_hash = index.lengths
length =
  case lengths_hash.size
  when 0
    nil
  when 1
    lengths_hash.values.first
  else
    lengths_hash
  end
```
I should probably try some of the corner cases to make sure they line up and don't get stuck in an endless migration loop due to aliasing, though. For example, I believe a single-column partial index can be requested either way in ActiveRecord:
```ruby
add_index :users, [:last_name], length: 10
```
or
```ruby
add_index :users, [:last_name], length: { last_name: 10 }
```
and also I wonder if ActiveRecord will allow the opposite on multi-column partial indexes, where they'll apply one length to all the columns?
```ruby
add_index :users, [:last_name, :first_name], length: 10
```

ttstarck (commented):
Looks fine, in the future I think it would be much easier to review this PR if there was a pre-factor PR, and then a PR to make the actual feature change.

lib/declare_schema/model/index_definition.rb
jebentier: Ah, I see. Thank you. Maybe we should utilize the same contract where `length` will always be a hash as well so that we don't have to have so many cases else where for handing different shapes of the length value. This could simply turn into:

```ruby
length = index.lengths
(length.nil? || length === Hash) or raise "Unexpected value #{length.inspect} provided for index lengths, expected a Hash"
```

ColinDKelley: That's an interesting compromise. I still like the simple Railsy configuration for single-column  indexes:
```ruby
index [:last_ name], length: 10
```
vs requiring this less-DRY form:
```ruby
index [:last_name], length: { last_name: 10 }
```
and of course there's symbol vs String aliasing too. :shakes_fist:

That all leads probably needing a normalize function. Something like
```
def normalize_index_length(columns, value)
  case value
  nil
    nil
  {}
    nil
  Hash
    value.transform_keys(&:to:s)
  Integer
    columns.each_with_object({}) { |column, hash| hash[column.to_s] = value }
  end
end
```
I could run the "desired state" config through that as well as what comes back from SQL in the `indexes` result.

ColinDKelley:
@ttstarck 
> in the future I think it would be much easier to review this PR if there was a pre-factor PR, and then a PR to make the actual feature change.

Yep, you're right. I've often done separate refactor PRs and that's always better, so I'll make sure to do it in the future.

In this case the refactor sneaked up a bit because I rediscovered [a refactor PR](https://github.com/Invoca/declare_schema/pull/157) that was approved in July but I'd forgotten to merge it. So, basically, it was compounding error. :doh:

Also, BTW, it's not that hard to create the refactor PR as a cherry-pick from the original. I could still do that here if you think it would help to see them separated.

lib/declare_schema/model/index_definition.rb
```diff
@@ -16,10 +18,12 @@ class IndexNameTooLongError < RuntimeError; end
 
       # Caller needs to pass either name or table_name. The table_name is not remembered; it is just used to compute the
       # default name if no name is given.
-      def initialize(columns, name: nil, table_name: nil, allow_equivalent: false, unique: false, where: nil)
+      def initialize(columns, table_name:, name: nil, allow_equivalent: false, unique: false, where: nil, length: nil)
+        @table_name = table_name
         @name = name || self.class.default_index_name(table_name, columns)
+        @name.to_s == 'index_adverts_on_Advert' and binding.pry
```
ttstarck, line 24: FYI this binding.pry is still here.

ttstarck (commented):
I'd rather review the changes separately, with the one refactor PR and the other to add  the index length feature.

ColinDKelley:
@ttstarck 
> I'd rather review the changes separately, with the one refactor PR and the other to add the index length feature.

Done. The prefactor is now in #164. The only exception is the README, which I didn't bother breaking out since it was pretty interleaved.

@jebentier, please re-review here when you get a chance

ColinDKelley:
@ttstarck Can you review the final changes here?
- I documented all the args and options to belongs_to. [Commit](https://github.com/Invoca/declare_schema/pull/162/commits/5693db90a4aed602c2bacb53216a2c7932e9cd5a)
- I added STDERR.puts warnings (not Deprecation.warn because that was getting suppressed) with [declare_schema]  prefix when belongs_to is passed unnecessary options as enumerated in the web PR. [Commit](https://github.com/Invoca/declare_schema/pull/162/commits/9a25a98326a09e7c6e185218c9303d948e6e3c78)

CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ Inspired by [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 Note: this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.4] - Unreleased
```
ttstarck, line 7: Nit: 
```suggestion
## [1.4.0] - Unreleased
```

ttstarck (approved):
Looks good!

PR #163, opened 2024-01-17
TECH-13772: fix MySQL 8 where charset "utf8" is normalized to "utf8mb3" and collation "utf8_..." is normalized to "utf8mb3_..."

ColinDKelley:
### Fixed
- Fix a MySQL 8 bug where MySQL 8+ renames charset 'utf8' to 'utf8mb3' and collation 'utf8_general_ci' to
  'utf8mb3_general_ci'.

This fix is applied in 3 places:
1. if specified globally in `DeclareSchema.default_charset/default_collation `;
2. if `charset` or `collation` is specified on a table;
3. if `charset` or `collation` is specified on a column

### DIFF
--- a/lib/declare_schema.rb
+++ b/lib/declare_schema.rb
@@ -64,22 +64,19 @@ module DeclareSchema
     end
 
     def normalize_charset(charset)
-      if mysql_version && mysql_version >= SEMVER_8
-        if charset == 'utf8'
-          'utf8mb3'
-        end
-      end || charset
+      if mysql_version && mysql_version >= SEMVER_8 && charset == 'utf8'
+        'utf8mb3'
+      else
+        charset
+      end
     end
 
     def normalize_collation(collation)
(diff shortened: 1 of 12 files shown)


lib/declare_schema.rb
```diff
@@ -21,6 +22,8 @@ class Boolean; end
     text:     String
   }.freeze
 
+  SEMVER_8 = Gem::Version.new('8.0.0').freeze
```
ColinDKelley, line 25: I'm cheating slightly to use `Gem::Version`. I've found it really handy for SemVers in general, since it implements comparison with any number of decimals.

jebentier (approved):
Changes look good and I like the version-aware portion to the change as well. I have a feeling it may come in handy in the future for version specific optimizations.

eddof13 (approved):

ColinDKelley:
@jebentier Can you review the final changes here? The build is green!

PR #164, opened 2024-01-18
TECH-13204/prefactor

ColinDKelley:
This is a prefactor branch for TECH-13204 #162


ColinDKelley:
@ttstarck @jebentier Pre-factors now appear here.

lib/declare_schema/model.rb
```diff
@@ -126,13 +114,35 @@ def belongs_to(name, scope = nil, **options)
...
 
-        index_options = {}
-        index_options[:name]   = options.delete(:index) if options.has_key?(:index)
-        index_options[:unique] = options.delete(:unique) if options.has_key?(:unique)
-        index_options[:allow_equivalent] = options.delete(:allow_equivalent) if options.has_key?(:allow_equivalent)
+        # index: true means create an index on the foreign key
+        # index: false means do not create an index on the foreign key
+        # index: { ... } means create an index on the foreign key with the given options
+        index_value = options.delete(:index)
+        if index_value != false || options.has_key?(:unique) || options.has_key?(:allow_equivalent)
```
ttstarck, line 121: `if index_value != false` looks strange, but is this to because we want to enter the block if `index_value` is nil?

lib/declare_schema/model.rb
```diff
@@ -126,13 +114,35 @@ def belongs_to(name, scope = nil, **options)
...
+          index_options = {} # truthy iff we want an index
+          case index_value
+          when String, Symbol
+            Kernel.warn("belongs_to index: 'name' is deprecated; use index: { name: 'name' } instead (in #{name})")
```
ttstarck, line 125: This looks like a new change, that is beyond refactoring.

Can you add this to the CHANGELOG showing that they've been deprecated?

lib/declare_schema/model.rb
```diff
@@ -126,13 +114,35 @@ def belongs_to(name, scope = nil, **options)
...
+            index_options[:name] = index_value.to_s
+          when false
+            raise ArgumentError, "belongs_to index: false contradicts others options #{options.inspect} (in #{name})"
```
ttstarck, line 128: These raises feel like breaking changes.

Are we sure we should include them in this PR?

lib/declare_schema/model/habtm_model_shim.rb
```diff
@@ -4,22 +4,25 @@ module DeclareSchema
...
+              [reflection.active_record.table_name, reflection.class_name.constantize.table_name],
+              connection: connection)
         end
       end
 
-      attr_reader :join_table, :foreign_keys, :parent_table_names
+      attr_reader :join_table, :foreign_keys, :parent_table_names, :connection
 
-      def initialize(join_table, foreign_keys, parent_table_names)
+      def initialize(join_table, foreign_keys, parent_table_names, connection:)
```
ttstarck, line 17: This feels like an internal change, but technically might be "breaking" as it now adds a new required argument to this class.

Thoughts?

lib/declare_schema/model/habtm_model_shim.rb
```diff
@@ -4,22 +4,25 @@ module DeclareSchema
...
         foreign_keys.is_a?(Array) && foreign_keys.size == 2 or
           raise ArgumentError, "foreign_keys must be <Array[2]>; got #{foreign_keys.inspect}"
         parent_table_names.is_a?(Array) && parent_table_names.size == 2 or
           raise ArgumentError, "parent_table_names must be <Array[2]>; got #{parent_table_names.inspect}"
         @join_table = join_table
         @foreign_keys = foreign_keys.sort # Rails requires these be in alphabetical order
         @parent_table_names = @foreign_keys == foreign_keys ? parent_table_names : parent_table_names.reverse # match the above sort
+        @connection = connection
```
ttstarck, line 25: What is the point of passing the connection into this class?

I'm not really seeing where / how its getting used

lib/declare_schema/model/index_definition.rb
```diff
@@ -104,12 +105,21 @@ def primary_key?
...
+
+      # Unique key for this object. Used for equality checking.
       def to_key
         @to_key ||= [name, *settings].freeze
       end
 
+      # The index settings for this object. Used for equivalence checking. Does not include the name.
       def settings
-        @settings ||= [columns, unique, where].freeze
+        @settings ||= [columns, options.except(:name)].freeze
```
ttstarck, line 122: This looks to have changed the format that this is returning right?
`unique` and `where` were strings, and `options.except(:name)` is a hash?

Is this expected?

lib/declare_schema/model.rb
```diff
@@ -95,7 +83,7 @@ def declare_field(name, type, *args, **options)
         _add_serialize_for_field(name, type, options)
         _add_formatting_for_field(name, type)
         _add_validations_for_field(name, type, args, options)
-        _add_index_for_field(name, args, options)
+        _add_index_for_field(name, args, **options)
```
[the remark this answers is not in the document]
ColinDKelley: The method now takes `**options`, so the calling code needs `**options` too, right?

But I might not be understanding your point?

lib/declare_schema/model.rb
ColinDKelley: Yeah, that is really confusing. It might not be right still.

If it's `false`, we can still enter the `if` for other reasons. But then we'll `raise` to indicate that `false` is contradictory.

But if it's `nil`, you're right that that will enter the `if` even if nothing else is set. And the `case` will do nothing. None of the code will do anything except set `index_options = {}`. Which is then treated as truthy below... which seems like it would then call `index`. Hmm....

ColinDKelley: Aha, I think I see how this is subtly broken. The base branch is slightly different, because it does:
```ruby
index_options[:name]   = options.delete(:index) if options.has_key?(:index)
```
So if `:index` isn't in options at all, it won't be copied across into `index_options`.

I think I need to check `has_key?` rather than `!= false`. And then clear it out with `delete` inside the `if`.

lib/declare_schema/model/habtm_model_shim.rb
ColinDKelley: This is an internal class only (not documented in the public interface) so I think it's fine.

lib/declare_schema/model/index_definition.rb
ColinDKelley: Yes, the `settings` just need to be unique. It wasn't necessary to flatten them into strings.

And in the past there had been bugs where an option was added but didn't get added to the settings or key. So I wanted them to track whatever's in `options`.

lib/declare_schema/model.rb
ColinDKelley: Update: I put the code back closer to what it was, without any raise on contradiction. I put it back to where`index: false` takes precedence over any other sub-settings for indexes.

lib/declare_schema/model.rb
ColinDKelley: This change should only have been in the feature branch. So I dropped it from the prefactor.

(It was already in the CHANGELOG over there.)

lib/declare_schema/model.rb
ColinDKelley: You're right, this belonged only in the feature branch. And I changed it to a warning over there.

lib/declare_schema/model/habtm_model_shim.rb
ColinDKelley: This is an internal-only class... but it turns out that was not needed any more, so I've dropped it. (`HabtmModelShim` needs to be polymorphic with `Model`, and when I added this, there was a `connection` method on `Model`. That's gone now.)

lib/declare_schema/model/habtm_model_shim.rb
ColinDKelley: It was strictly for the `attr_reader :connection`. But good call to flag it because that was no longer used by the migrator.

lib/declare_schema/model.rb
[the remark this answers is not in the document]
ttstarck: Huh... I may have just missed that in the method definition. I swore I was the method definition as the following when I wrote the comment:
```
def _add_index_for_field(column_name, args, options)
```

I see now that it is defined with the `**options`

ttstarck (approved):
Looks good! Thanks for pulling this out into a separate PR

PR #166, opened 2024-01-22
TECH-13204/make default config lazy

ColinDKelley:
### Fixed
- Make `default_charset=` and `default_collation=` lazy so they don't use the database connection to check the
  MySQL version. Instead, that is checked the first time `default_charset` or `default_collation` is called.

### DIFF
--- a/lib/declare_schema.rb
+++ b/lib/declare_schema.rb
@@ -36,7 +36,7 @@ module DeclareSchema
 
   class << self
     attr_writer :mysql_version
-    attr_reader :default_charset, :default_collation, :default_text_limit, :default_string_limit, :default_null,
+    attr_reader :default_text_limit, :default_string_limit, :default_null,
                 :default_generate_foreign_keys, :default_generate_indexing, :db_migrate_command,
                 :max_index_and_constraint_name_length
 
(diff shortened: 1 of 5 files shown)


lib/declare_schema.rb
```diff
@@ -81,12 +81,22 @@ def normalize_collation(collation)
 
     def default_charset=(charset)
       charset.is_a?(String) or raise ArgumentError, "charset must be a string (got #{charset.inspect})"
-      @default_charset = normalize_charset(charset)
+      @default_charset_before_normalization = charset
+      @default_charset = nil
+    end
+
+    def default_charset
+      @default_charset || normalize_charset(@default_charset_before_normalization)
```
jebentier, line 89: ❓ Should this be an `||=`?

lib/declare_schema.rb
```diff
@@ -81,12 +81,22 @@ def normalize_collation(collation)
...
 
     def default_collation=(collation)
       collation.is_a?(String) or raise ArgumentError, "collation must be a string (got #{collation.inspect})"
-      @default_collation = normalize_collation(collation)
+      @default_collation_before_normalization = collation
+      @default_collation = nil
+    end
+
+    def default_collation
+      @default_collation || normalize_collation(@default_collation_before_normalization)
```
jebentier, line 99: ❓ Same question here.

lib/declare_schema.rb
ColinDKelley: Yes, good catch! Both fixed here: https://github.com/Invoca/declare_schema/pull/166/commits/2def468ac7d426a0d4c29df2b2a50063eabf36f9

lib/declare_schema.rb
ColinDKelley: Both fixed here: https://github.com/Invoca/declare_schema/pull/166/commits/2def468ac7d426a0d4c29df2b2a50063eabf36f9

jebentier (approved):

PR #169, opened 2024-05-12
Non production test against newer rails versions

jebentier:
The purpose of this work is to get full Rails 7+ support up to the most recent minor version release of Ruby. In this PR you'll find:

* Updates to the build pipeline to include more Appraisal configurations and more ruby versions being tested
* Updates to tests to account for slight variations in migration text output between versions
* Updates to internals to account for new functionality available in newer versions of Ruby

This PR has not gone to lengths to ensure that everything that was added in Rails 7.1 (which includes direct support for composite primary keys) is accounted for. This just makes sure that existing functionality works.


Rakefile
```diff
@@ -29,7 +29,7 @@ namespace "test" do
   task :prepare_testapp, :force do |_t, args|
     if args.force || !File.directory?(TESTAPP_PATH)
       FileUtils.remove_entry_secure(TESTAPP_PATH, true)
-      sh %(#{BIN} new #{TESTAPP_PATH} --skip-wizard --skip-bundle)
+      sh %(#{BIN} new #{TESTAPP_PATH} --skip-wizard --skip-bundle --api)
```
jebentier, line 32: 📓 Added this to make bootstrapping easier since we don't need to test with a frontend capable rails application, this trims down the chances of configuration errors cropping up.

lib/declare_schema/model.rb
```diff
@@ -263,9 +263,13 @@ def _infer_fk_limit(foreign_key_column, reflection)
       # unlike the `primary_key` method, DOES NOT query the database to find the actual primary key in use right now
       # if no explicit primary key set, returns the _default_declared_primary_key
       def _declared_primary_key
-        if defined?(@primary_key)
+        if !defined?(@primary_key) ||
+           (ActiveSupport.version >= Gem::Version.new('7.1.0') &&
+             @primary_key == ActiveRecord::AttributeMethods::PrimaryKey::ClassMethods::PRIMARY_KEY_NOT_SET)
```
jebentier, line 268: 📓 In Rails 7.1+ they're now using a constant `BasicObject` instance as a reference to the primary key not being set yet, so this additional check is now necessary.

lib/generators/declare_schema/migration/migration_generator.rb
```diff
@@ -117,6 +106,36 @@ def migrations_pending?
...
+    def load_migrations
+      if ActiveSupport.version >= Gem::Version.new('7.1.0')
+        ActiveRecord::MigrationContext.new(
+          ActiveRecord::Migrator.migrations_paths,
+          ActiveRecord::Base.connection.schema_migration,
+          ActiveRecord::Base.connection.internal_metadata
+        ).migrations
+      elsif ActiveSupport.version >= Gem::Version.new('6.0.0')
+        ActiveRecord::MigrationContext.new(ActiveRecord::Migrator.migrations_paths, ActiveRecord::SchemaMigration).migrations
+      else # Rails 5.2 and earlier
```
jebentier, line 118: ❓ Should we go head and rip out Rails 5 support from the gem and make this a major version bump instead?

spec/lib/declare_schema/generator_spec.rb
```diff
@@ -19,13 +19,23 @@ class Alpha::Beta < #{active_record_base_class}
...
 
-    expect_model_definition_to_eq('alpha', <<~EOS)
-      module Alpha
-        def self.table_name_prefix
-          'alpha_'
+    if ActiveSupport::VERSION::MAJOR >= 7
+      expect_model_definition_to_eq('alpha', <<~EOS)
+        module Alpha
+          def self.table_name_prefix
+            "alpha_"
```
jebentier, line 26: 😞 It's unfortunate that we need to separate out blocks of tests like this to account for quotation usage differences. Maybe in the future we can find a better way to test that doesn't care which quotes are used.

spec/lib/declare_schema/generator_spec.rb
```diff
@@ -88,7 +98,11 @@ class Alpha::BetaTest < ActiveSupport::TestCase
     expect(File.exist?('db/schema.rb')).to be_truthy
 
     if defined?(SQLite3)
-      expect(File.exist?("db/development.sqlite3") || File.exist?("db/test.sqlite3")).to be_truthy
+      if ActiveSupport.version >= Gem::Version.new('7.1.0')
+        expect(File.exist?("storage/development.sqlite3") || File.exist?("storage/test.sqlite3")).to be_truthy
```
jebentier, line 102: 📓 Rails 7.1 moved the default location of sqlite databases into the `storage/` directory to make gitignoring easier.

spec/lib/declare_schema/migration_generator_spec.rb
```diff
@@ -30,7 +30,9 @@
     end
   end
   let(:datetime_precision) do
-    if defined?(Mysql2)
+    if ActiveSupport::VERSION::MAJOR >= 7
```
jebentier, line 33: 📓 Rails 7+ now sets datetime precision to 6 instead of 0 for mysql or blank for sqlite.

spec/lib/declare_schema/migration_generator_spec.rb
```diff
@@ -860,21 +862,39 @@ class Creative < ActiveRecord::Base
...
+            create_table :advertisers_creatives, primary_key: [:advertiser_id, :creative_id]#{create_table_charset_and_collation} do |t|
+              t.integer :advertiser_id, limit: 8, null: false
+              t.integer :creative_id, limit: 8, null: false
+            end
+            create_table :creatives, id: :bigint#{create_table_charset_and_collation} do |t|
+              t.string :url, limit: 500, null: false#{charset_and_collation}
+            end
+            create_table :advertisers, id: :bigint#{create_table_charset_and_collation} do |t|
+              t.string :name, limit: 250, null: false#{charset_and_collation}
+            end
```
jebentier, line 876: 📓 This only difference on this block and the below block is that the create table ordering is different, seems that rails 7+ on ruby 3.1+ orders by most recent declared models first.

spec/spec_helper.rb
```diff
@@ -39,11 +39,15 @@ def migrate(renames = {})
   end
 
   def nuke_model_class(klass)
-    ActiveSupport::DescendantsTracker.instance_eval do
-      direct_descendants = class_variable_get('@@direct_descendants')
-      direct_descendants[ActiveRecord::Base] = direct_descendants[ActiveRecord::Base].to_a.reject { |descendant| descendant == klass }
-      if defined?(ApplicationRecord)
-        direct_descendants[ApplicationRecord] = direct_descendants[ApplicationRecord].to_a.reject { |descendant| descendant == klass }
+    if ActiveSupport.version >= Gem::Version.new('7.1.0') || (ActiveSupport.version >= Gem::Version.new('7.0.0') && Class.method_defined?(:subclasses))
+      ActiveSupport::DescendantsTracker.clear([klass])
```
jebentier, line 43: 📓 With Ruby 3.1 the `subclasses` method introduced to be able to track descendants, and Rails 7+ started to take advantage of this, allowing for a sleek new way to clear a class from the descendant tracking. The internals now just track and "exclude" list and this `clear` method appends to that list.

lib/generators/declare_schema/migration/migration_generator.rb
ttstarck: Yeah, I'd say so. I think in general its common practice to only support 2 major versions at a time. This seems reasonable to me that we'd also only support two major versions.

jebentier: Done 189372263b0e6a2fae3a6b51fa5e992507a7012c

yinonrousso (approved):
👏

ttstarck (approved):
Looks good!

PR #171, opened 2024-08-15
Non production improve dev and test environment

jebentier:
The core of this PR is updating how we build/run tests to no longer rely on what gems are available to be loaded so that it's easier to test against multiple different configurations locally. Instead it utilizes the active connection configuration to understand how the test app is configured so know if it should be running mysql or sqlite specific tasks.

I started out doing this, because I want to add full postgresql support to declare schema to enhance it's functionality, but adding the necessary testing infrastructure would have been a large lift.

Added benefit of this work is we are now testing and green against Rails 7.2, but would like to wait on releasing that change until we've had a more dedicated look at it.


ColinDKelley (approved):
Very nice improvement!

I like that it gets us one step closer to adding Postgres testing, too.

PR #172, opened 2024-08-18
Add support for PostgreSQL databases

### LINKED ISSUE #154: Index names are not unique

jebentier:
This PR is adding deliberate support to `declare_schema` for PostgreSQL databases. The majority of changes required were around testing support and getting the Github action updated, but the explicit gaps that were filled to work with Postgres are the following:

* Switched logic used to retrieve `foreign_keys` from the existing database schema to use built-in ActiveRecord method to make sure no matter what adapter is used, they can still be retrieved
* Fixed SchemaDumper logic to use the appropriate dumper given the current adapter
* Started using the configured adapter over the connection class name for understanding what type of database we're connected to (this allows the checks to evaluate properly before a connection can be established)
* Added a `.devcontainer` configuration to help with local testing against multiple databases at once
* Always use the long index name for Postgres connections due to the global uniqueness requirement
* Using `bigserial` as the default primary key type for tables in `postgresql`
* Use public method for getting cast type for column

Fixes #154


lib/declare_schema/model/index_definition.rb
```diff
@@ -64,6 +64,8 @@ def for_table(table_name, ignore_indexes, connection)
         end
 
         def default_index_name(table_name, columns)
+          return long_index_name(table_name, columns) if DeclareSchema.current_adapter == 'postgresql'
+
```
ColinDKelley, line 68: Rather than early `return`, can you make the rest of this method a new private method, say `longest_valid_index_name`? Then this method will become
```ruby
def default_index_name(table_name, columns)
  if DeclareSchema.current_adapter == 'postgresql'
    long_index_name(table_name, columns)
  else
    longest_valid_index_name(table_name, columns)
  end
end
```

README.md
```diff
@@ -115,6 +115,8 @@ The following `index` options are supported:
 - `length` (integer or hash) - The partial index length(s). If an integer is provided, it is used as the length for all columns. If a hash is provided, it is used to specify the length for individual columns, where the column names are given as `Symbol` hash keys.
 - `where` (string) - The subset index predicate.
 
+**Note:** The `DeclareSchema.max_index_and_constraint_name_length` setting is ignored when using `PostgreSQL` since that database does not have a limit on the length of index names, and requires globally unique index names.
+
```
ColinDKelley, line 119: How about adding an explicit section for "Tested Databases"? That would list that it's tested with:
- Sqlite
- Postgres
- MySQL

It might be cool to even mention the major versions we're testing with, like 5.7 and 8.0 for MySQL.

lib/declare_schema/model/index_definition.rb
jebentier: Get idea, done 057d3c26a57ca6a24f6a54c6d4e97b31f91330c3

CONTRIBUTING.md
```diff
@@ -0,0 +1,46 @@
+# Contributing to ContextualLogger
```
ColinDKelley, line 1: I think this should be `Contributing to DeclareSchema`. Or maybe just make it `Contributing` so it's ready for future copy/paste?

README.md
```diff
@@ -372,19 +392,18 @@ class Comment < ActiveRecord::Base
...
-  $ gem install declare_schema
-```
-or add it to your `bundler` Gemfile:
-```
-  gem 'declare_schema'
-```
-## Testing
+Contributions to this project are always welcome. Please thoroughly read our [Contribution Guidelines](CONTRIBUTING.md) before starting any work.
+
+### Local Development
```
ColinDKelley, line 399: 💯 Thanks for filling this in.

ColinDKelley (approved):
Looks great.

Just one copy/paste correction to make.

CONTRIBUTING.md
jebentier: Thanks for catching this, fixed 1e67bb195e516981f55287cd7d5c9f74b9ef586e

garvank:
Thanks for the work on this Colin & James! Love the gem 🎉

PR #173, opened 2024-10-08
TECH-16372: drop/add index if settings change (that order is important in case index name is identical)

ColinDKelley:
I ran into this bug while working  https://invoca.atlassian.net/browse/WEB-7889 . I know I'd seen it before. This time I tracked it down.

### Changed
- Drop/add index if its settings change (`unique:`, `length:`, etc.)
- Drop indexes before adding, in case any have identical names (so we don't get a SQL exception about the duplicate name)
- Same for foreign key constraints
- Address CVE warning with `bundle update rexml`

### DIFF
--- a/lib/generators/declare_schema/migration/migrator.rb
+++ b/lib/generators/declare_schema/migration/migrator.rb
@@ -285,10 +285,10 @@ module Generators
                                 ColumnRename
                                 ColumnChange
                                   PrimaryKeyChange
-                                  IndexAdd
-                                    ForeignKeyAdd
-                                    ForeignKeyRemove
                                   IndexRemove
+                                  IndexAdd
+                                    ForeignKeyRemove
+                                    ForeignKeyAdd
                                 ColumnRemove
                               TableRemove ]
 
@@ -464,7 +464,7 @@ module Generators
           indexes_to_drop = existing_indexes_without_primary_key - model_indexes_without_primary_key
           indexes_to_add = model_indexes_without_primary_key - existing_indexes_without_primary_key
 
-          renamed_indexes_to_drop, renamed_indexes_to_add = index_changes_due_to_column_renames(indexes_to_drop, indexes_to_add, to_rename)
+          renamed_indexes_to_drop, renamed_indexes_to_add = index_changes_solely_due_to_column_renames(indexes_to_drop, indexes_to_add, to_rename)
 
           drop_indexes = (indexes_to_drop - renamed_indexes_to_drop).map do |i|
             ::DeclareSchema::SchemaChange::IndexRemove.new(new_table_name, i.columns, **i.options)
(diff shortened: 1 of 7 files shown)


lib/generators/declare_schema/migration/migrator.rb
```diff
@@ -285,10 +285,10 @@ def generate
...
                                 ColumnChange
                                   PrimaryKeyChange
-                                  IndexAdd
-                                    ForeignKeyAdd
-                                    ForeignKeyRemove
-                                  IndexRemove
+                                  ForeignKeyRemove
+                                    IndexRemove
+                                    IndexAdd
+                                  ForeignKeyAdd
```
ColinDKelley, line 291: I realized that Remove-then-Add is necessary because indexes/constraints are named and SQL enforces that the names be unique. Also foreign keys depend on indexes so it makes the most sense to remove the foreign keys before removing indexes.

lib/generators/declare_schema/migration/migrator.rb
```diff
@@ -478,17 +478,25 @@ def change_indexes(model, old_table_name, to_rename)
           [Array(change_primary_key) + drop_indexes + add_indexes]
         end
 
-        def index_changes_due_to_column_renames(indexes_to_drop, indexes_to_add, to_rename)
-          indexes_to_drop.each_with_object([[], []]) do |index_to_drop, (renamed_indexes_to_drop, renamed_indexes_to_add)|
+        # @return [Array<Array<IndexDefinition>>] pair of arrays of indexes that changed solely due to column renames...not any settings changes.
```
ColinDKelley, line 481: The contract here was the source of the bug. The calling code needs to know index changes that _only_ happened because of column renames, since those are implied by renaming the columns and don't need a separate schema change.

lib/generators/declare_schema/migration/migrator.rb
```diff
@@ -478,17 +478,25 @@ def change_indexes(model, old_table_name, to_rename)
...
+        # Note: if the column order changed, that is considered a change of settings--not solely a column rename--so it will not be returned
+        def index_changes_solely_due_to_column_renames(indexes_to_drop, indexes_to_add, to_rename)
+          renamed_indexes_to_drop = []
+          renamed_indexes_to_add = []
+
+          indexes_to_drop.each do |index_to_drop|
             renamed_columns = index_to_drop.columns.map do |column|
               to_rename.fetch(column, column)
-            end.sort
```
ColinDKelley, line 485: This was a misfeature/bug I think. Column order is significant in SQL indexes. There wasn't a spec for this, so I added one.

lib/generators/declare_schema/migration/migrator.rb
```diff
@@ -478,17 +478,25 @@ def change_indexes(model, old_table_name, to_rename)
...
+            end
 
-            if (index_to_add = indexes_to_add.find { |index_to_add| renamed_columns == index_to_add.columns.sort })
+            if (index_to_add = indexes_to_add.find { |index_to_add| renamed_columns == index_to_add.columns }) &&
+                index_to_add.settings == index_to_drop.settings
```
ColinDKelley, line 493: Here's the actually bug fix that the `settings` have to be the same. (`settings` is like `options` but doesn't include the name.)

spec/lib/declare_schema/migration_generator_spec.rb
```diff
@@ -585,21 +586,64 @@ class Advert < ActiveRecord::Base # rubocop:disable Lint/ConstantDefinitionInBlo
 
       Advert.index_definitions.delete_if { |spec| spec.fields == ["title"] }
 
+      ActiveRecord::Migration.class_eval(up)
+
+      # You can migrate an index to have slightly different settings
```
ColinDKelley, line 591: This is the spec for the main bug.

spec/lib/declare_schema/migration_generator_spec.rb
```diff
@@ -585,21 +586,64 @@ class Advert < ActiveRecord::Base # rubocop:disable Lint/ConstantDefinitionInBlo
...
+          add_index :adverts, [:title], name: :my_index, unique: true, length: { title: 5 }
           add_index :adverts, [:title, :category_id], name: :index_adverts_on_title_and_category_id
         EOS
       )
 
+      ActiveRecord::Migration.class_eval(up)
+
       Advert.index_definitions.delete_if { |spec| spec.fields == ["title", "category_id"] }
 
+      # You can change the column order of an index
```
ColinDKelley, line 633: This is the spec for the secondary bug. Just swapping the order of the columns now causes the index to be dropped then added below.

spec/lib/declare_schema/migration_generator_spec.rb
```diff
@@ -365,7 +365,7 @@ class Advert < ActiveRecord::Base # rubocop:disable Lint/ConstantDefinitionInBlo
       )
 
       Advert.field_specs.delete(:category_id)
-      Advert.index_definitions.delete_if { |spec| spec.fields == ["category_id"] }
+      Advert.index_definitions.clear
```
jebentier, line 368: 👏

jebentier (approved):
Changes look good! Thank you for taking care of this and cleaning up the bug.

PR #174, opened 2024-10-10
TECH-16372: allow_equivalent: true and separate ignore_equivalent_definitions: false

ColinDKelley:
When integrating `declare_schema` into web, a lot of false migrations (in Warehouse) got kicked off because the `allow_equivalent: false` which was also partially true in this gem. The problem was overloading that for two purposes which are now separated here:

- `allow_equivalent: true` This allows an index of a different name but same settings to suffice. Setting it to `false` will cause the old index to be dropped and the new one added if the name is different (which is very common since we've changed the default names so many times!).
- `ignore_equivalent_definitions: false` This makes index definition idempotent. That wouldn't come up when indexes are declared by hand, but Warehouse needs it in 1 or 2 places because it's code generating the definitions.

### DIFF
--- a/lib/declare_schema/model/index_definition.rb
+++ b/lib/declare_schema/model/index_definition.rb
@@ -16,9 +16,9 @@ module DeclareSchema
 
       PRIMARY_KEY_NAME = "PRIMARY"
 
-      def initialize(columns, table_name:, name: nil, allow_equivalent: false, unique: false, where: nil, length: nil)
+      def initialize(columns, table_name:, name: nil, allow_equivalent: true, unique: false, where: nil, length: nil)
         @table_name = table_name
-        @name = (name || self.class.default_index_name(table_name, columns)).to_s
+        @name = name&.to_s || self.class.default_index_name(table_name, columns)
         @columns = Array.wrap(columns).map(&:to_s)
         @allow_equivalent = allow_equivalent
         unique.in?([false, true]) or raise ArgumentError, "unique must be true or false: got #{unique.inspect}"
--- a/lib/declare_schema/model/index_definition.rb
+++ b/lib/declare_schema/model/index_definition.rb
@@ -8,7 +8,7 @@ module DeclareSchema
       include Comparable
 
       OPTIONS = [:name, :unique, :where, :length].freeze
-      attr_reader :columns, :explicit_name, :table_name, *OPTIONS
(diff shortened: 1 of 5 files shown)


lib/declare_schema/model/index_definition.rb
```diff
@@ -8,19 +8,19 @@ class IndexDefinition
       include Comparable
 
       OPTIONS = [:name, :unique, :where, :length].freeze
-      attr_reader :columns, :explicit_name, :table_name, *OPTIONS
+      attr_reader :columns, :allow_equivalent, :table_name, *OPTIONS
```
ColinDKelley, line 11: This was goofy that we didn't just remember the option under its own name.

jebentier:
I'm trying to understand the Warehouse issue a little better, is it making false migrations because index names in the schema don't match what we're setting in the ruby code (misalignment) or is it because we're duplicating definitions through inheritance structures which is causing duplicate index definitions.

Based on the definition provided above about `allow_equivalent: true`, does that mean (and I know this is madness) if I put duplicate index settings definitions in a class, it might flip flop between the two? Example:

```ruby
class Webhook
  index [:owner_id], unique: true, name: :webhook_owner_id, allow_equivalent: true
  index [:owner_id], unique: true, name: :webhooks_owner_id, allow_equivalent: true
end
```

Or because the names are different, will these be seen as two separate settings?

ColinDKelley:
> I'm trying to understand the Warehouse issue a little better

The biggest surprise was the index uniqueness was previously ignored for equivalence! Once that was fixed over here, Warehouse generated a lot of migrations, primarily because of [this change I had made](https://github.com/Invoca/web/blame/master/app/models/warehouse/base/aggregate.rb#L34) from `unique: true` to `unique: false`. That needed to go back to how it was. But also it was important to separate the options for `allow_equivalent:` and `ignore_equivalent_definitions:`. Formerly the first was used for both meanings...and there were bugs over here about the implementation and default for that which are fixed in this PR.

I have a web PR coming up now that will show the Warehouse change along with a bunch of deprecation warning fixes.

jebentier (approved):
Looks good!

ttstarck:
@ColinDKelley we need to update declare_schema in web with a fix, but this minor version has broken all of web. Can you help us get web bumped?

PR #176, opened 2024-12-10
TECH-16946: fix bug on new model with polymorphic belongs_to

ColinDKelley:
### Fixed
- Fixed bug where a new model with `belongs_to :owner, polymorphic: true` would cause
  a "Mysql2::Error: Table '<new table>' doesn't exist:" exception when generating a migration.


ttstarck (approved):
Looks great! Thanks for digging into this

gmcgilvray (approved):
Looks good!

PR #177, opened 2025-01-10
add support for rails 8.0

ttstarck:
Add support for Rails 8.0

Note: Ruby 3.4 seems to work, however there are tests in here that are failing so I'm leaving out testing this for now.


lib/declare_schema/dsl.rb
```diff
@@ -1,7 +1,5 @@
 # frozen_string_literal: true
 
-require 'active_support/proxy_object'
```
ttstarck, line 3: We're no longer using this at all, and it is removed in Rails 8.0

lib/declare_schema/field_declaration_dsl.rb
```diff
@@ -1,7 +1,5 @@
 # frozen_string_literal: true
 
-require 'active_support/proxy_object'
```
ttstarck, line 3: We're no longer using this at all, and it is removed in Rails 8.0

Gemfile
```diff
@@ -26,3 +26,9 @@ gem 'yard'
 gem 'mysql2',  '~> 0.5'
 gem 'pg',      '~> 1.1'
 gem 'sqlite3', '~> 1.4'
+
+gem "base64", "~> 0.2.0"
+
+gem "mutex_m", "~> 0.3.0"
+
+gem "bigdecimal", "~> 3.1"
```
ttstarck, line 34: This is for Ruby 3.4 but tests are failing due to how `Hash#inspect` changed so i've left fixing that for later

ishakun (approved):

PR #182, opened 2025-04-08
Tech 18196 fix bug in value validations and expand functionality of timestamps dsl

jebentier:
### Changed
- The `timestamps` DSL method to create `created_at` and `updated_at` columns now defaults to `null: false` for `datetime` columns
- The `timestamps` DSL method to allow additional options to be passed to the `datetime` fields

### Fixed
- Fixed a bug where `#validate` methods on core object classes with required arguments were cuasing model validations to fail

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,14 @@ Inspired by [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 Note: this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [2.3.3] - Unreleased
+### Changed
+- The `timestamps` DSL method to create `created_at` and `updated_at` columns now defaults to `null: false` for `datetime` columns
+- The `timestamps` DSL method to allow additional options to be passed to the `datetime` fields
+
+### Fixed
+- Fixed a bug where `#validate` methods on core object classes with required arguments were cuasing model validations to fail
+
 ## [2.3.2] - 2025-02-21
 ### Fixed
 - Removed require of `activesupport/proxy_object` which is removed in Rails 8.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ Note: this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0
 - The `timestamps` DSL method to allow additional options to be passed to the `datetime` fields
 
 ### Fixed
-- Fixed a bug where `#validate` methods on core object classes with required arguments were cuasing model validations to fail
+- Fixed a bug where `#validate` methods on core object classes with required arguments was causing model validations to fail
 
 ## [2.3.2] - 2025-02-21
 ### Fixed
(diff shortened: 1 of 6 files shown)


lib/declare_schema/dsl.rb
```diff
@@ -17,9 +17,9 @@ def initialize(model, **options)
 
     attr_reader :model
 
-    def timestamps
-      field(:created_at, :datetime, null: true)
-      field(:updated_at, :datetime, null: true)
+    def timestamps(**options)
+      field(:created_at, :datetime, null: false, **options)
+      field(:updated_at, :datetime, null: false, **options)
     end
```
ttstarck, line 23: Is this a breaking change? Will this cause all `created_at` and `updated_at` fields to need migrations?

jebentier: 🤔 It could be considered breaking, but there is the ability to have previous functionality by using `timestamps null: true` to override the new default. I think that's enough to not require the breaking change bump, but not strongly opposed.

ttstarck: >It could be considered breaking but there is the ability to have previous functionality by using timestamps null: true to override the new default

This is definitely breaking since it requires code changes, its just good practice to provide an easy way of going back to old behavior 😅. This should also be a minor version at the minimum since this is adding new functionality too.

I don't know if other people are really using this gem besides invoca but if there are, we should definitely make this a breaking change.

Or we can split this into two PRs:
1. Minor version bump: Add new functionality of passing options to timestamp fields.
2. Major version bump: Change default value of `null: true` to `null: false`

ttstarck: If you don't think there are other users of our gem, I'm fine with doing this as a minor version bump

jebentier: That's a good point. [Dependents graph](https://github.com/Invoca/declare_schema/network/dependents) says only 4 other users, so minor seems good for this

jebentier: But, I agree we should follow best practice here. I'll cut it as major. fd8668ea578748cf02fb002ae8d5885acae4d762

CHANGELOG.md
```diff
@@ -4,6 +4,14 @@ Inspired by [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 Note: this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [3.0.0] - Unreleased
+### Changed
+- The `timestamps` DSL method to create `created_at` and `updated_at` columns now defaults to `null: false` for `datetime` columns
+- The `timestamps` DSL method to allow additional options to be passed to the `datetime` fields
+
+### Fixed
+- Fixed a bug where `#validate` methods on core object classes with required arguments were cuasing model validations to fail
```
ttstarck, line 13:
```suggestion
- Fixed a bug where `#validate` methods on core object classes with required arguments was causing model validations to fail
```

ttstarck (approved):
Just one final fix for a typo in the changelog

PR #183, opened 2026-05-02
WEB-8347: support non-default primary key types for belongs_to and HABTM foreign keys

ColinDKelley:
## Summary

WEB-8347: support non-`:bigint` primary key types for `belongs_to` and `has_and_belongs_to_many` foreign keys.

`declare_schema` previously assumed every primary key was `:bigint` (or whatever the live DB column happened to be after the fact). When a model used a non-default PK type — e.g. `declare_schema id: { type: :string, limit: 36 }` — the foreign keys generated for `belongs_to` and HABTM associations did not mirror that type, producing migrations that MySQL would reject as incompatible.

The fix is to defer the FK's type decision from `belongs_to` time (when the parent class might not be loaded yet — dependency-cycle hazard) to migration-generation time (after `eager_load!`), via a `&resolver` callback installed on a `DeferredFieldSpec`. The migration generator invokes the resolver, which mirrors the parent's `_primary_key_field_spec` (type, limit, charset, collation, etc.) and reconciles against the live DB column when it differs only in `:limit`.

## What changed

**Production**
- `lib/declare_schema/model.rb`
  - `_infer_foreign_key_field_spec` (non-polymorphic path) installs a `resolver:` callback on a `:bigint` placeholder instead of touching `reflection.klass` eagerly.
  - New `_resolve_belongs_to_foreign_key_field_spec` runs at migration generation time. Mirrors the parent's PK spec via `klass._primary_key_field_spec.foreign_key_field_spec(...)` and reconciles `:limit` against the live DB column (handles legacy tables where declared bigint defaults disagree with on-disk INT(4)).
  - `_primary_key_field_spec_from_table_options` now handles the bare-Symbol form (`declare_schema id: :integer`) in addition to the Hash form. The Hash/Symbol parsing is encapsulated in a small private helper `_parse_pk_table_options`.
- `lib/generators/declare_schema/migration/migrator.rb`
  - Generator's pre-generation pass invokes each `:resolver` and swaps in the FieldSpec it returns. The contract is intentionally minimal: take the placeholder, return the final FieldSpec.
  - Removes `fk_field_options`, the post-hoc DB-column lookup that was redundant after the resolver lands and was actively wrong for non-default PK types (it would overwrite a correct `limit:` with whatever the DB happened to have).
- `lib/declare_schema/model/habtm_model_shim.rb`
  - Falls back to a `:bigint` `FieldSpec` when a HABTM parent isn't a `declare_schema` model (instead of `NoMethodError` on `_foreign_key_field_spec`).

**Specs**
- `spec/lib/declare_schema/migration_generator_spec.rb`
  - Three new specs covering: deferred lookup (no `NameError` when child references parent before it's loaded), `belongs_to` mirroring `:string, limit: 36` PK, HABTM mirroring two parents with different non-default PK types.
- `spec/lib/declare_schema/model/habtm_model_shim_spec.rb`
  - Fix pre-existing constructor signature drift (`HabtmModelShim.new` was changed earlier on this branch from `(join_table, foreign_keys, parent_table_names, connection:)` to `(join_table, parents, connection:)` but the spec subject hadn't been updated).

## Test plan

- [x] `bundle exec rake test:prepare_testapp[mysql,true] && bundle exec rake test:all` → **253 examples, 0 failures, 0 pending**
- [x] `bundle exec rake test:prepare_testapp[sqlite3,true] && bundle exec rake test:all` → **253 examples, 0 failures, 8 pending** (8 are MySQL-only)
- [ ] Reviewer: confirm semantics for the live-DB reconciliation in `_resolve_belongs_to_foreign_key_field_spec` — we override `:limit` only when type matches and limit differs; intentional type changes (e.g. bigint → string) are preserved.
- [ ] Reviewer: sanity-check the `:resolver` callback name; we deliberately avoided "pre_migration:" since it's not a callback that runs before a migration is *applied*, but a resolver invoked during migration *generation*.


lib/declare_schema/extensions/active_record/fields_declaration.rb
```diff
@@ -3,7 +3,6 @@
 require 'active_record'
 require 'declare_schema/dsl'
 require 'declare_schema/model'
-require 'declare_schema/field_declaration_dsl'
```
ColinDKelley, line 6: Turns out this was dead code! There's a newer `Dsl` class.

lib/declare_schema/model/field_spec.rb
```diff
@@ -47,10 +47,6 @@ def round_up_mysql_text_limit(limit)
       end
 
       def initialize(model, name, type, position: 0, **options)
-        _declared_primary_key = model._declared_primary_key
-
-        name.to_s == _declared_primary_key and raise ArgumentError, "you may not provide a field spec for the primary key #{name.inspect}"
-
```
ColinDKelley, line 53: This assertion moved up a layer.

lib/generators/declare_schema/migration/migrator.rb
```diff
@@ -189,11 +196,7 @@ def generate
           models, db_tables = models_and_tables
           models_by_table_name = {}
           models.each do |m|
-            m.try(:field_specs)&.each do |_name, field_spec|
-              if (pre_migration = field_spec.options.delete(:pre_migration))
-                pre_migration.call(field_spec)
-              end
-            end
+            m.try(:field_specs)&.transform_values!(&:resolve)
```
ColinDKelley, line 199: This is the main change of the PR. Every model gets resolved (if not already) just before generating the migration. Resolution happens just in time here so we can see the shape of the type and options of the primary key we're pointing to. For example, it could be a `type: :binary, { limit: 16 }`. Previously we'd only adapt the integer limit between 4 and 8.

lib/generators/declare_schema/migration/migrator.rb
```diff
@@ -204,8 +207,8 @@ def generate
             end
           end
           # generate shims for HABTM models
-          habtm_tables.each do |name, reflections|
-            models_by_table_name[name] = ::DeclareSchema::Model::HabtmModelShim.from_reflection(reflections.first)
+          habtm_tables.each do |name, reflection|
+            models_by_table_name[name] = ::DeclareSchema::Model::HabtmModelShim.from_reflection(reflection)
```
ColinDKelley, line 211: We weren't looking beyond the first, so why hold the list at all?

lib/generators/declare_schema/migration/migrator.rb
```diff
@@ -546,22 +553,6 @@ def foreign_key_changes_due_to_column_renames(fks_to_drop, fks_to_add, to_rename
           end
         end
 
-        def fk_field_options(model, field_name)
```
ColinDKelley, line 549: This is the old code that fixed up the `limit` but didn't know how to have an entirely different foreign key `type`, like `binary(16)`.

lib/declare_schema.rb
```diff
@@ -162,19 +162,24 @@ def deprecator
     end
 
     def current_adapter(model_class = ActiveRecord::Base)
-      if Rails::VERSION::MAJOR >= 6.1
-        model_class.connection_db_config.adapter
-      else
-        model_class.connection_config[:adapter]
-      end
+      model_class.connection_db_config.adapter
```
ColinDKelley, line 165: Dropped support of Rails 6.

jebentier (approved):

PR #187, opened 2026-07-01
OCTO-892: Document the field macro in the README

ColinDKelley:
## Summary
- Documents the `field` DSL macro (the primitive underlying every `t.<type>` shorthand, `timestamps`, and `optimistic_lock`), including its `:required`/`:unique` positional flags.
- Documents the underlying `declare_field` class method, which can be called outside of a `declare_schema do ... end` block (e.g. from a shared concern), with an equivalent example for each style.
- Bumps the gem to `4.0.3` (docs-only release) and updates the CHANGELOG.

## Test plan
- [x] Reviewed rendered README section for accuracy against `lib/declare_schema/dsl.rb` and `lib/declare_schema/model.rb`
- [x] `bundle install` succeeds and picks up `declare_schema (4.0.3)`

Jira: OCTO-892

Made with [Cursor](https://cursor.com)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ Inspired by [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 Note: this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [4.0.3] - 2026-07-01
+### Changed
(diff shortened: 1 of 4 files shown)


gmcgilvray (approved):
Thanks for doing this! 👍 though I did not dig through the docs for correctness, but appreciate the tag and pointing out this functionality.

jebentier (approved):

